PR #5, opened 2019-07-10
Add note on multithreading configuration when installing Chez Scheme

abailly:


edwinb:
Good idea, thanks! Hopefully this will be enough.

PR #9, opened 2019-07-12
install target depend on all target

diakopter:


edwinb:
I'm using the 'install' target to build and install the libraries without having to build the executable too. The 'idris2' target can be slow to build. So this would be a bit annoying for me. Is this necessary, or is there another way that doesn't break my usual workflow?

melted:
I guess one could have

```
install: all install-libs

install-libs:
   <current stuff>
```

One reason to have the dependency is that install is a very common target, so people expect it to work like usual.

diakopter:
edited to support both the install-libs workflow and the typical install behavior

edwinb:
Seems reasonable, thanks!

PR #16, opened 2019-07-15
Add --clean option for ipkg

clayrat:
Since the `fRemove` FFI call only works on empty folders, I first construct a trie of all modules and then delete them recursively by traversing the trie depth-first.


edwinb:
Seems useful, thanks! This is slightly less trivial to merge than some of the others because I need to check my Idris is up to date, but I'll catch up tomorrow...

PR #18, opened 2019-07-16
allow running IDE mode on different network interfaces

symbiont-arnaud-bailly:
This is a first attempt at making it possible to run the IDE mode server on a non-localhost interface, in order to allow it to run inside a container for example, or on a remote machine. I have several questions:

* Why is there 2 different `Main.idr` files with similar code? Seems like `Idris.REPL.Main` is unused
* How can I write tests for this? Do I need to enhance the test runner to include something to spin up a main with socket-mode and connect to it?
* How can I make it possible to drop the argument to `--ide-mode-socket` to be backward compatible?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@
 idris2
 runtests
 
+docs/_build/
+
 libs/**/build
 tests/**/output
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -35,13 +35,14 @@ extensions = [
     'sphinx.ext.todo',
 #    'sphinx.ext.pngmath', # imgmath is not supported on readthedocs.
     'sphinx.ext.ifconfig',
+    'recommonmark'
 ]
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
 
 # The suffix of source filenames.
-source_suffix = '.rst'
+source_suffix = ['.rst','.md']
 
 # The encoding of source files.
 #source_encoding = 'utf-8-sig'


edwinb:
Thanks for this! I'll try to check it in the next day or two (until we have CI it's a bit more fiddly to check over this sort of thing I'm afraid...)

In the meantime, the answer to your first question about why there's two `Main.idr` files is probably... oops! I will have copied over some bits from Blodwen without thinking, and `Idris.REPL.Main` should just be deleted.

It's probably not a bad idea to have a test runner which can talk to the socket mode. I haven't really thought about it - the tests as they stand are mostly about trying things with the core language. There's plenty more that's worth testing beyond that.

symbiont-arnaud-bailly:
Thanks for the answer @edwinb. One of the interesting outcome of this feature is that it would make it easier to have a CI working using containers, which is what most CI providers do those days.

I will leave the supernumerary `Idris.REPL.Main` alone atm and possibly delete it in another PR, just to not confuse things. And I was about to try writing some tests involving the ide socket, so will add them to the PR.

Any suggestions as to my last question? In essence I would to be able to write either:
```
idris2 --ide-socket-mode foo:12312 --other-arg file.idr
```
or 
```
idris2 --ide-socket-mode --other-arg file.idr
```
eg. make the explicit `host:port` argument to the socket optional. I tried using the current command-line parsing infrastructure but failed. Any suggestion on which direction to investigate?

edwinb:
Hmm, is the command line parsing thing because there's currently no way to have a flag with an optional argument? "Infrastructure" is quite a grand word for what it is at the moment, but I think that should be achievable...

symbiont-arnaud-bailly:
Yes. I could not find a way to have an optional argument. I tried modifying the type family `ActType` to something like:
```
ActType : List String -> Type
ActType [] = List CLOpt
ActType (a :: as) = 
  if "?" `isPrefixOf` a
  then Maybe String -> ActType as
  else String -> ActType as
```
but it did not "work".

I don't have any idea to make that work "out-of-the-box" without changing the parser.

edwinb:
I've just tried this, and it froze on the second ideMode test. I'm not sure where to start on debugging that one...

symbiont-arnaud-bailly:
@edwinb Could it be the case that you have some local firewall activated? I am running it locally and got:
```
typedd-book/chapter12: success
ideMode/ideMode001: success
ideMode/ideMode002: success
Found Chez Scheme at /usr/local/bin/scheme
chez/chez001: success
chez/chez002: success
chez/chez003: success
chez/chez004: success
chez/chez005: success
chez/chez006: success
chez/chez007: success
140/140 tests successful
```

Of course, those kind of tests are inherently brittle as they require a fixed port to be free which might not always be the case. Maybe we can just remove this test?

abailly:
I can't make this test run reliably using the current test infrastructure, it would need more involved job control logic and not running in a bash script. It's easy to test manually thought.
@edwinb Are you happy to merge this as is, knowing that the `--ide-mode-socket` now requires a a `host:mode` argument or do you want to me to turn this argument optional?

edwinb:
I think in general, if a PR makes useful progress without being perfect, I'd rather err on the side of merging, as long as it doesn't break anything. So if you're happy for me to merge, I'll do that. Thanks!

abailly:
Thanks. I would rather make it possible to have optional arguments, that would be cleaner and ensure it won't break anyone 's flow. Even though there aren't that many people using it right now but still.

abailly:
@edwinb Could you close this PR? I will provide another PR on another branch.

PR #22, opened 2019-07-17
Use fgetc instead of getc

PeterHajdu:
On openbsd getc is a macro that expects its parameter to be FILE*
whereas the generated code provides void*.


edwinb:
Ah yes, I think someone did this for Blodwen in the, er, limbo period while I was hacking on Idris 2. Thanks!

PR #27, opened 2019-07-18
Add Data.Morphisms to base

shmish111:


edwinb:
I guess this is a direct port from Idris 1, yes? Thanks!

shmish111:
Apart from the lack of idiom brackets and `!` yes it is. Also `neutral _ = neutral` that was used in Idris 1 didn't work although I can't really understand how it worked in Idris 1.

edwinb:
I wouldn't be surprised if it shouldn't have worked in Idris 1, to be honest ;)

Issue #26, opened 2019-07-18
Undefined name


nicolabotta:
The program
```
module issues.UndefinedName

||| An isomorphism between two types                                                                                      
public export
record Iso a b where
  constructor MkIso
  to : a -> b
  from : b -> a
  toFrom : (y : b) -> to (from y) = y
  fromTo : (x : a) -> from (to x) = x

||| Disjunction is commutative
public export
eitherComm : Iso (Either a b) (Either b a)
eitherComm = MkIso mirror mirror mirrorMirror mirrorMirror where
  {-
  mirror : Either a' b' -> Either b' a'
  mirror (Left x) = Right x
  mirror (Right y) = Left y
  ---}
  mirrorMirror : (e : Either a' b') -> mirror (mirror e) = e
  mirrorMirror (Left x) = Refl
  mirrorMirror (Right x) = Refl
```
type checks in Idris 1 but not in Idris 2. 
```
issues.UndefinedName> :r
1/1: Building issues.UndefinedName (issues/UndefinedName.idr)
issues/UndefinedName.idr:23:40--23:58:While processing right hand side of issues.UndefinedName.eitherComm at issues/UndefinedName.idr:17:1--26:1:
While processing type of 1161:mirrorMirror at issues/UndefinedName.idr:23:3--24:3:
Undefined name mirror
issues.UndefinedName>
```
I have no problem with the user having to provide an implementation of `mirror`, I was actually a bit surprised that Idris 1 can do without it! I am reporting the issue just in case Idris 2 is expected to behave as Idris 1 in this case.

edwinb:
Is 'mirror' in the Idris 1 prelude? That's the only reasonable explanation
I can come up with... I don't think I've added anything to base for Either
yet.

edwinb:
Okay, I decided it was probably better to check than to guess :).

My guess is indeed correct, though. I'm going to close this as expected behaviour, and I'll add a `Data.Either` module soon unless someone else gets to it first. (Adding things from the Idris prelude into the Idris2 base libraries is part of the wish list in CONTRIBUTING.md.)

PR #28, opened 2019-07-18
Inclusion of Either within Base.

jfdm:
Straightforward port of Either from Idris to Idris2.


edwinb:
Thanks! Looks relevant to #26

PR #31, opened 2019-07-18
add talk video link

diakopter:


edwinb:
Seems worth noting, thanks!

PR #35, opened 2019-07-20
Add CI

Theodus:
This PR adds a Travis CI configuration. Currently this is using a set of Docker containers on a minimal Travis image to avoid building both Idris and Idris2 during each CI run.

### DIFF
--- a/.ci-dockerfiles/ci-base.dockerfile
+++ b/.ci-dockerfiles/ci-base.dockerfile
@@ -1,0 +1,19 @@
+from haskell:8.2
+
+run apt-get update && apt-get install -y \
+  clang lld-4.0 \
+  curl wget \
+  chicken-bin racket \
+  libx11-dev uuid-dev \
+  libncurses5-dev
+
+run wget https://github.com/cisco/ChezScheme/archive/v9.5.2.tar.gz && \
+  tar xvzf v9.5.2.tar.gz && \
+  cd ChezScheme-9.5.2 && \
+  ./configure --threads && \
+  make -j install
+
+run cabal update && cabal install -j --with-ld=ld.lld-4.0 idris-1.3.2
+
+env IDRIS_CC=clang
+cmd idris
--- a/.ci-dockerfiles/ci-run.dockerfile
+++ b/.ci-dockerfiles/ci-run.dockerfile
@@ -1,0 +1,7 @@
+from theodus/idris2-ci-base:latest
+
+copy . /Idris2-dev
+workdir /Idris2-dev
+run make idris2
+run make libs
+run make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+dist: bionic
+
+services: docker
+
+language: minimal
+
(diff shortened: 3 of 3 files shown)


Theodus:
The CI seems to be running into memory problems and getting killed when running `make idris2 libs`. Here's the [log](https://circleci.com/gh/Theodus/Idris2/4?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). I'll give it another go with clang.

edwinb:
Thanks for doing this! This'll be really helpful once it's up and running.

I have found that the generated C builds much quicker with clang. Although I'm surprised if this takes more memory than type checking the Idris code.

Theodus:
[Same problem when using clang](https://circleci.com/gh/Theodus/Idris2/5#queue-placeholder/containers/0). I'm not sure how else to reduce the memory consumption of this step, so unless someone else has some suggestions I might try this again in a few days using Travis CI.

.ci-dockerfiles/idris-4e70401.dockerfile
```diff
@@ -0,0 +1,11 @@
+from haskell:8.2
+
+run apt-get update && apt-get install -y libncurses5-dev
+
+run git clone https://github.com/idris-lang/Idris-dev && \
+  cd Idris-dev && \
+  git checkout 4e70401 && \
+  cabal update && \
+  make -j install
```
chshersh: Try using `-j1` instead of `j`. We had the same problems with Circle CI. It runs out of memory when tries to build with cabal in parallel with multiple threads.

```suggestion
  make -j1 install
```

Theodus:
Update: I have a working version with Travis, but 30 minutes are spent installing Idris(1). So I'll need to sort that out before it's usable.

kristleifur:
> Update: I have a working version with Travis, but 30 minutes are spent installing Idris(1). So I'll need to sort that out before it's usable.

Can we solve that by building it only once? I.e. making that idris1 binary that has been built into a tagged image and pushing it to Dockerhub? Then it could be run as a `FROM` stage – or optionally the binary/libs copied in via `COPY --from` if the idris1 image is not desired as a base image for an idris2.

Theodus:
I was thinking of something similar, by using the `docker cp` command to pull the binary out of the container. Travis sadly doesn't allow you to start from a docker image like CircleCI does.

Theodus:
I have an example of this running successfully here: https://travis-ci.org/Theodus/Idris2/builds/562759067

Attempts to extract the Idris install from the base container have failed, and this seems to be a cleaner solution anyway. @edwinb, if you prefer to host the base CI image in another Docker Hub account let me know. Otherwise, we can leave that for another time.

kristleifur:
This looks very clean and practical, and reusable!

edwinb:
That's great, thank you! I'm happy to do whatever works :). Do you think this is ready to merge now?

Theodus:
Yup, I think it's ready

edwinb:
Great! This'll be a huge help, thanks.

PR #37, opened 2019-07-20
Fix primitive function for acos

christian-public:
Small fix.


edwinb:
Oops, well spotted :). I guess this sort of thing is likely to happen when adding things by copy and paste so I'm glad someone has had a closer look...

PR #38, opened 2019-07-21
Document base for exp and log functions

christian-public:
Might be helpful for people writing code generators.


edwinb:
Looks a good idea, thanks. I mostly copied this from Idris 1's behaviour.

PR #40, opened 2019-07-23
Add missing double quotes

ska80:


edwinb:
Hmm, I wonder how the tests didn't catch this. I'd better do something about that. Thanks!

PR #43, opened 2019-07-24
port base.Data.List from Idris1

clayrat:
I've separated `Elem` into a submodule to avoid making `Data.List` too bloated.


edwinb:
Looks good, thanks

PR #45, opened 2019-07-25
First draft porting network code from Idris1 to Idris2

symbiont-arnaud-bailly:
This PR is probably not polished enough to warrant merge but I would appreciate some feedback on this work and possibly some advice on how to make it usable. It works as demonstrated by the `Echo.idr` file and the `lib-tests` command which is a very basic integration test demonstrating echoing a client from a server.
I still have a couple of questions:
* How does this work with installation under `~/.idris2/` directory? The problem is how the compiler/chez backend will know where to look for the libraries to load? AFAIK this can be solved either by tinkering with the OS's library paths loading or by explicitly giving the path to the library file. The latter would lead to possibly non-portable and non-moveable installation directory which is annoying in the long run.
* How does the end-user should use such packages which needs to load external libraries? It seems that compiler directives are not "imported" which means end-users would have to explicitly add `%cg chez libidris_net` explicitly each time they want to use the network.

The porting was mostly straightforward, replacing `foreign ...` calls with `cCall` and taking care of the return values and parameters to be primitives.


libs/network/Echo.idr
```diff
@@ -0,0 +1,78 @@
...
+import Network.Socket.Raw
+
+%cg chez libidris_net.dylib
+
+runServer : IO (Either String (Port, ThreadID))
+runServer = do
+  osock <- socket AF_INET Stream 0
+  case osock of
+    Left fail => pure (Left $ "Failed to open socket: " ++ show fail)
+    Right sock => do
```
clayrat: These functions would probably look a bit nicer with pattern matching bindings, i.e.,
```
Right sock <- socket AF_INET Stream 0
  | Left fail => pure (Left $ "Failed to open socket: " ++ show fail)
...
```

symbiont-arnaud-bailly: I always forget this... Thanks for the tip, will update the code.

edwinb:
I've been wondering how to deal with the chez backend loading foreign code, specifically teaching it where to look. I don't have a good answer to this yet but this library will need one! At least if they're in some standard place in the idris system path, the chez code will know where to look. Also it's slightly annoying that different systems use different extensions for dynamic libraries by default, but maybe this isn't a big deal since as far as I can tell chez doesn't care what the file is called. 

CG directives are supposed to be imported, so clearly something is going wrong if they're not...

symbiont-arnaud-bailly:
Thanks for the feedback @edwinb.
Would installing those foreign libraries into `~/.idris2/lib/` make sense? Then we could instruct the chez backend to load those from this "well-known" path. I think I can make the needed change and see how it looks like?

edwinb:
I think looking in ~/.idris2/lib makes sense, at least for the moment. Longer term we'll probably need to do something else though (e.g. if someone wants to distribute the generated chez without Idris being available).

edwinb:
I was just reading the chez scheme docs on distributing programs (https://cisco.github.io/ChezScheme/csug9.5/use.html#./use:h8) - maybe an ugly-but-portable solution would be to have the chez CG output a directory containing the generated scheme and necessary support libraries, along with a one line script which invokes the scheme program?

I hope that we won't be using chez forever, and we'll move to a dedicated native code CG at some point, but while we have it we should probably try to use it most effectively.

abailly:
That sounds like a good tactical solution. I still need to fix the CI execution, looks like dynamic loader resolution mechanism is different between mac and Linux as one can expect.

shmish111:
This is _the_ library to define how the FFI will work. In the future we'll probably want this library to work with all of; chicken, WASI, node.js, future native RTS as well as working on Windows, OSX, Linux, other crazy platforms and finally be able to build it in sandboxes as used in nix and bazel (without make) and possibly other build systems, hopefully including one written in Idris!

All this and we'll want to be able to distribute it as a transient dependency.

IMO this is the most important library that will exist, I also note that the network library in Haskell has caused me more problems than all other libraries combined 😅

Not sure if this comment is helpful but I just want to explain how important I think this is.

abailly:
>  I also note that the network library in Haskell has caused me more problems than all other libraries combined 

How so? I seldom have had issues with this library, at least directly. But I have used it mostly indirectly...

shmish111:
I've had quite a few problems building on OSX with both stack and nix. Always with network as a deeply buried transient library. I think it's usually things that happen when upgrading OSX and/or the network library itself. It's happened on a few separate occasions over about 2 years.

abailly:
@edwinb I merged your recent changes on library loading but it does not solve the issue of different extensions/names for different OSes. Here are the rules I am aware of and I tried to implement in reverted commit 9e7d79c, given base name is `idris_net`, linking at compile time is `-lidris_net`:
* Linux: library should be named `libidris_net.so` (or possibly `libidris_net.so.X` where `X` is a version number)
* Mac OS: library should be named `libidris_net.dylib` 
* Windows: library should be named `idris_net.dll`

I am pretty sure this is a pretty simplistic view, and it is possible to explicitly load dynamic libraries which don't follow the standard rules of the OS, but that was enough to pass the build and the basic tests I have on Mac OS and Linux.

Should I try to enhance the existing code to take of OS-specific naming? 
Alternatively, we could also decide that foreign libraries should be referenced using an explicit filename with extension that is normalised to `.so`. This is what is suggested in the chez user guide: http://cisco.github.io/ChezScheme/csug9.5/foreign.html#./foreign:h6

shmish111:
Is it possible to use header files instead of so/dylib/dll files in the directive? Then use c style `-l` flags?

abailly:
@shmish111 It's definitely possible. My goal with this PR was to get something off-the-ground and then iterate from there while the FFI evolves and allow other stuff to be build up on a pure Idris2 interface with a well-known semantic (eg. sockets). How would you like things to look like?

shmish111:
Yeah, I appreciate the idea of getting things actually working first. My main concern with the FFI is having things in the Idris code that are non-portable. You came across the first and biggest one of the dylib extension, perhaps this could be solved by using header files instead. That would force Idris to have done sort of path search algorithm which I think it's a good thing, as long as it includes the ability to add explicit paths not just system and expected paths. For example in Bazel it would be easy to copy header files to the build sandbox but the dylib files are stored in a temporary sandbox path that you would need to explicitly tell the Idris linker about.

I am looking forward to building your library when I'm back from holiday then I can give you some more concrete issues.

shmish111:
I was just thinking about the case of having multiple backends and I thought of one possible solution.

You could have multiple source directories, one for common code, one for chez backend, one for JVM backend etc. In your build tool you could work out a way to only include the source directories for a specific platform.

This would mean that the library you want to use would have to support the backend you want to use but in the worst case you could fork it yourself.

This doesn't really work for platforms though I don't think, you would end up with a mess of tens of directories 😀

edwinb:
@abailly Do you think this is okay to merge now? Even if it's not completely polished, as long as the basics work it'll be good to have it in master.

edwinb:
For the FFI thing, since I would hope a lot of backends would be able to talk via C but in different ways, it probably makes sense to have directives for giving headers and libraries then let each code generator deal with it in their preferred ways.

abailly:
What's obviously tricky is that there are different stages to take into consideration: The type checker/Idris compiler phase, the Idris -> Backend code generation phase, the backend -> executable phase, the runtime phase... 

In the present situation, with native C libraries, the scheme backend(s) are really the ones responsible for the dynamic loading of the library using the OS primitives (`dlopen` for Unices, `LoadLibrary` for Windows, maybe others I am not aware of...) so it seems to me Idris should delegate that part: There's no real "linking" process happening like in a C/C++ program when doing `-lxxx`.

abailly:
@edwinb That would be great. At least it compiles and pass "all" the tests :)

shmish111:
The reason I mention the JVM backend is because it specifically is unlikely to use C code. It is a very rare thing these days to use the FFI in Java and for something like networking there are already great libraries.

shmish111:
It makes sense to delegate just as long as there is a way to pass the required flags from the command line or ipkg file.

abailly:
@shmish111 Not sure GitHub is the perfect medium do have this discussion :) Maybe we can switch to slack?

I mean I am super-happy to discuss all this but the GitHub interface is clumsy

edwinb:
Great, thanks for all this, I'll merge in a bit - I'm slightly scared because it's a big one so I'm going to test it manually first!

edwinb:
I'm getting a surprising error when I build this:

    cp idris_net.so libidris_net.so # to make C linker happy
    cc -o network-tests -L. -I. -lidris_net test.c
    /tmp/ccHPU4Sp.o: In function `test_sockaddr_port_returns_random_port_when_bind_port_is_0':
    test.c:(.text+0x18): undefined reference to `idrnet_socket'
    test.c:(.text+0x61): undefined reference to `idrnet_bind'
    test.c:(.text+0x93): undefined reference to `idrnet_sockaddr_port'

Given that it works on Travis, this is probably something odd about my system. If I add idris_net.o explicitly to the `cc` command, it's fine. Similarly if I `export CC=clang` it's fine (which probably explains the difference). Is it that gcc doesn't look for dynamic libraries with -l?

I'm not sure of the right thing to do here...

abailly:
Good question. I thought the linking was pretty much universal but
apparently is not... You are running this on OS X ?

Le mer. 31 juil. 2019 à 13:46, Edwin Brady <notifications@github.com> a
écrit :

edwinb:
I'm on Ubuntu 18.something - the difference appears to be between how gcc and clang handle things. I can make it work for myself no problem, I'm just a bit worried about difficulty of installing in general.

abailly:
I think I made a mistake in the `Makefile`:

```
	$(CC) -o network-tests -L. -I.  -l$(LIBNAME) test.c
```

should be

```
	$(CC) -o network-tests -L. -I. test.c -l$(LIBNAME)
```

according to this https://stackoverflow.com/questions/22426574/gcc-undefined-reference-to

abailly:
I agree about the potential difficulties of installing native libraries. I basically c/p the Idris1 Makefile and adapted it, but writing a universally portable Makefile for native C code is definitely not trivial. CMake alleviates a great deal of this issue at the expense of additional complexity of maintaining yet another build tool...

abailly:
Also I did not test that on Windows, need to spin up a VM somewhere and tackle that, but maybe in a later PR?

edwinb:
Great, looks like everything is working now. Thanks for getting this all working!

PR #47, opened 2019-07-26
Check idris version before building

symbiont-arnaud-bailly:
The actual version might actually not be correct...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,18 @@
 PREFIX ?= ${HOME}/.idris2
+IDRIS_VERSION=1.3.2-git:9549d9cb9
 export IDRIS2_PATH = ${CURDIR}/libs/prelude/build:${CURDIR}/libs/base/build
 export IDRIS2_DATA = ${CURDIR}/support
 
 -include custom.mk
 
-.PHONY: ttimp idris2 prelude test base clean lib_clean
+.PHONY: ttimp idris2 prelude test base clean lib_clean check_version
 
 all: idris2 libs test
 
-idris2: src/YafflePaths.idr
+check_version:
+	@if [ `idris --version` != "$(IDRIS_VERSION)" ]; then echo "Wrong idris version, expected $(IDRIS_VERSION)"; exit 1; fi
+
+idris2: src/YafflePaths.idr check_version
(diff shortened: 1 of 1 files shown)


clayrat:
This doesn't seem future-proof, e.g. what if 1.3.3 comes out?

symbiont-arnaud-bailly:
You change the reference? The point is that one needs a specific version of Idris to build Idris2 so better pin it down in the Makefile and ensure someone building Idris2 will get an explicit error if his Idris installation is incorrect. Having a version range and a way to compare them would be better in the long run.

symbiont-arnaud-bailly:
But sure, having a version range and a way to compare them would be better.

melted:
The release builds of Idris, such as the binary builds or the tarballs on hackage doesn't feature the git commit part. It's way too narrow a test anyway, pinning it to a specific commit makes it impossible to work on Idris 1 without hassle.

symbiont-arnaud-bailly:
I totally agree, hence my comment:

> The actual version might actually not be correct...

What's the correct version that should be checked?

melted:
the version would differ depending on whether it was built from git or from hackage (or using a binary build), the git commit part is omitted on release builds. You could just check if it starts with 1.3.2, because we know that all 1.3.2 works.

melted:
Try this 

```
$(expr "$(idris --version)" : '1.3.[2-9].*')
```

melted:
really, that should be `\.` as `.` matches anything. But of course `.` is something  :-)

Makefile
```diff
@@ -1,14 +1,20 @@
 PREFIX ?= ${HOME}/.idris2
+IDRIS_VERSION := $(shell idris --version)
+VALID_IDRIS_VERSION_REGEXP = "1.3.3.*"
```
melted: 1.3.2

edwinb:
I'd hope that Idris2 would still build with some hypothetical future version of Idris 1, but since there isn't one yet we can worry about that later... also at least in my case, all the changes I've been making to Idris 1 for a while have been to support the development of Idris 2...

Anyway, as it is, this should hopefully help with explaining installation problems!

PR #51, opened 2019-07-26
fix implicit/brace block ambiguity in records

### LINKED ISSUE #46: Implicit record fields
# Steps to Reproduce
```
module Temp

import Data.Vect

record Foo where
  constructor MkFoo
  filler : Int
  {size : Nat}
  foos  : Vect size String

isEmpty : Foo -> Bool
isEmpty f = size f == 0
```

# Expected Behavior
This compiles and works as expected both with and without the `filler` field in Idris 1.3.

# Observed Behavior
Without `filler`:
```
Temp.idr:8:3--8:3:Parse error: Expected end of input (next tokens: [identifier foos, symbol :, identifier Vect, identifier size, String, identifier isEmpty, symbol :, identifier Foo, symbol ->, identifier Bool])
```

With `filler`:
```
Temp.idr:12:13--12:20:While processing right hand side of Temp.isEmpty at Temp.idr:12:1--13:1:
Undefined name size
```

clayrat:
Resolves the ambiguity between 
```
record Foo where 
  constructor MkFoo
  <brace block>
```
and
```
record Foo where
  constructor MkFoo
  { field : Ty }
```
by moving constructor parsing into the block. 
Partially resolves #46.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 *~
 *.ibc
+*.ttc
+*.ttm
 
 idris2
 runtests
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -1151,13 +1151,16 @@ recordDecl fname indents
          n <- name
          params <- many (ifaceParam fname indents)
          keyword "where"
-         dc <- option Nothing (do exactIdent "constructor"
-                                  n <- name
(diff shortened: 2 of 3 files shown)


edwinb:
I've changed the name generation for field lookups too so I think #46 is completely resolved now

PR #53, opened 2019-07-27
Only display IDE port when listening

kevinboulain:
[idris-mode](https://github.com/idris-hackers/idris-mode) is unhappy with Idris 2.

I get this error because the port is displayed before listening on the socket:
```
Debugger entered--Lisp error: (file-error "make client process failed" "Connection refused" :name "Idris IDE support" :buffer "*idris-connection*" :host "127.0.0.1" :service 38398 :nowait nil :tls-parameters nil)
  make-network-process(:name "Idris IDE support" :buffer "*idris-connection*" :host "127.0.0.1" :service 38398 :nowait nil :tls-parameters nil)
  open-network-stream("Idris IDE support" "*idris-connection*" "127.0.0.1" 38398)
...
```

Still doesn't make it usable though, but that's a first simple step :)

### DIFF
--- a/src/Idris/IDEMode/REPL.idr
+++ b/src/Idris/IDEMode/REPL.idr
@@ -54,7 +54,6 @@ initIDESocketFile p = do
       putStrLn "Failed to open socket"
       exit 1
     Right sock => do
-      putStrLn (show p)
       res <- bind sock (Just (Hostname "localhost")) p
       if res /= 0 
       then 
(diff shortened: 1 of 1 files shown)


edwinb:
I'm not really sure why it's printing the port at all to be honest... maybe it was a debugging aid? Anyway, all progress towards getting Idris 2 working with emacs is good, thanks!

kevinboulain:
I guess that's how clients can know the port when unspecified (so that you can leave the application choose a random port without any race)? Idris does the same:
```
idris --ide-mode-socket
49916
^C
idris --ide-mode-socket
49935
^C
```

PR #54, opened 2019-07-27
Make it a little easier to debug IDE mode

kevinboulain:
Following #53, I get an unhelpful (for me, because I don't know much about the protocol itself) "Unrecognised command". Fortunately `SExp` implements `Show`, so should we display the command?

Example from Emacs/idris-mode:
```
00003d(:return (:error "Unrecognised command: (:version 3)" ()) 2)
```


edwinb:
Seems reasonable, thanks

PR #60, opened 2019-07-28
Run individual tests (Fixes #23)

### LINKED ISSUE #23: Add a way of running tests individually
Many tests are failing when I "make test", so I want to run a single test and manually compare its result with contents of the `expected` file. So I try this:

```
~/Desktop/Idris2-master/tests/idris2/basic011 $ sh ./run ../../../idris2
(implicit):1:1--1:1:Prelude not found
(implicit):1:1--1:1:Prelude not found
(implicit):1:1--1:1:Prelude not found
```

So I must be doing something wrong? Is this the correct way to try to run an individual test? Would be nice if there was some documentation for this.

christian-public:
I added an `--only` flag to `runtests` that allows you to run only a subset of the tests. The tests are filtered using `String.isInfixOf`, so it does a partial match. Also, before filtering the tests the code will prepend the directory where the tests are located, which allows filters like `--only idris2`, `--only ttimp/basic001` etc.

I modified the Makefiles to propagate an `only` variable. Thus, to run all `chez` tests one could write: `make test only=chez`.

`make test` does still run all tests, though `make` will report that it runs the following commands:
```
make -C tests only=
../runtests ../../../idris2 --only 
```
It works because `runtests` expects another argument after `--only`, which it is not given. I tried adding an `if` around each command, but then the whole `if` was printed. I could also silence these commands by prepending `@` to the commands. Is this preferable?

Should I document this feature somewhere?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ clean-libs:
 
 test:
 	idris --build tests.ipkg
-	make -C tests only=$(only)
+	@make -C tests only=$(only)
 
 install: all install-exec install-libs
 
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ clean-libs:
 
 test:
 	idris --build tests.ipkg
-	make -C tests
+	make -C tests only=$(only)
 
 install: all install-exec install-libs
 
--- a/tests/Main.idr
+++ b/tests/Main.idr
@@ -70,10 +70,10 @@ fail err
     = do putStrLn err
          exitWith (ExitFailure 1)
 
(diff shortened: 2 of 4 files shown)


edwinb:
That's great, thanks! I suppose the best place to document it would be via a README in the tests directory, which doesn't currently exist.

christian-public:
I added a bare-bones README. I decided to silence the commands in the Makefiles: The presence of `only` in the test output could potentially cause confusion.

edwinb:
Excellent, looks good, thanks for this

PR #61, opened 2019-07-29
document %cg code generation directive for DLL loading

symbiont-arnaud-bailly:
This is a result of some experiments while working on #45


edwinb:
Thanks for this! It's one of many things I haven't got around to documenting properly yet. It's not *quite* right though - cg directives are passed directly to the backend for them to interpret exactly as they like. Different back ends may need very different things, after all. What you've described here, therefore, is chez specific.

The chicken backend just inserts the directive into the generated code unchanged.

symbiont-arnaud-bailly:
Agree. I will update the documentation to reflect that.

edwinb:
Great, thanks for the updates.

PR #66, opened 2019-07-30
check noprelude option when starting up without loading a file

abailly:


edwinb:
Great, thanks for the fix!

PR #69, opened 2019-07-31
Improve RTD config.

jfdm:
+ Sphinx can read CommonMark formatted files, so let us allow that capability.
+ Improve .gitignore to ignore sphinx build directories.


edwinb:
Cheers

PR #70, opened 2019-07-31
Align the IPKG format more with the previous Idris implementation. 

jfdm:
Align fields that would be useful for package management and documentation needs.

This PR also adds user documentation for Idris Packages.
This is an initial copy and refinement of the existing documentation for packaging.
This commit also starts the Idris2 reference documentation, in which we can start documenting the language proper. Maybe we should include the contents of `Notes` here, but that is a question for later.


edwinb:
There seems to be a missing test here...

jfdm:
For some reason I decided that the contents of expected should include timing information. That is obviously not a good thing to have in 'golden file testing'. I've changed it to *just* the output of Idris2.

edwinb:
Great, thanks for the fix!

PR #72, opened 2019-08-03
Print expected and given output when test fails.

jfdm:
Useful when looking to diagnose tests that fail.


edwinb:
Seems useful, thanks. I wonder if a diff might be better, but this is better than what we had (i.e. nothing...)

PR #73, opened 2019-08-03
Add output file flag (-o) to CLI

christian-public:
Adds `--output`/`-o` to Idris 2's CLI. This flag will compile and generate output file in the current directory.

`idris2 -o <outfile> Test.idr` is the same as performing `:c <outfile> main` from the REPL. Note that this flag expects the entrypoint function to be named `main`. This is similar to what Idris 1 does.

Have tested that it works with Chez and Racket (but not Chicken).

### DIFF
--- a/src/Idris/CommandLine.idr
+++ b/src/Idris/CommandLine.idr
@@ -23,6 +23,8 @@ data CLOpt
   =
    ||| Only typecheck the given file
   CheckOnly |
+   ||| The output file from the code generator
+  OutputFile String |
    ||| Execute a given function after checking the source file
   ExecFn String |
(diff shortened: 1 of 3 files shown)


edwinb:
Thanks!

How do you specify the CG for it though? Just looking at the code it seems to take the default. Perhaps also there should be a test, or some of the existing chez tests could be updated, but I think it's still safe to merge as it is.

christian-public:
Thanks for merging!

The `--cg` flag still works when using the `-o`/`--output` flag.

I am sorry for not providing a test in this pull request. I will fix this soon :-)

PR #75, opened 2019-08-06
Configurable ide-mode-socket REPL

abailly:
This replaces #18 

Why is this PR relevant? 

1. Because it makes it possible to package idris2 in a docker container, and run it in IDE mode while still being accessible from outside the container, eg. the running host on which a user runs its editor. This way Idris2 can be distributed as a docker image hence removing the need to run its own binaries
2. Because it makes it easy to run 2 or more Idris2 IDE REPL on the same host


abailly:
I went down the route of simplicity: add a different `--ide-mode-socket-with` option that takes an argument, instead of trying to tweak the command-line parser to have an optional argument to `--ide-mode-socket`. I could not find a better way to handle that issue.

edwinb:
This seems like a good thing to do, but unfortunately since I've been away and slow to merge things, there's ended up being conflicts. Would you mind fixing them up and then I'll merge? It doesn't look too tricky but I'd like to be sure I'm not breaking anything. Thanks.

abailly:
Sure thing, will do asap !

Le jeu. 29 août 2019 à 12:00, Edwin Brady <notifications@github.com> a
écrit :

edwinb:
Sorry I'm slow again, start of term is holding me up from doing interesting things... thanks for the fix!

PR #76, opened 2019-08-07
Ignore build artifacts

ska80:


abailly:
Might be better to put all build artefacts into `build/` directory?

ska80:
> Might be better to put all build artifacts into `build/` directory?

Yes, it could be a better solution as the 'build' directory already ignored.

edwinb:
Seems reasonable, thanks. I guess we do need to think a bit more about how to handle build artefacts of foreign code.

PR #80, opened 2019-08-11
Ignore empty lines in REPL

timsueberkrueb:
This ignores empty lines in the REPL and doesn't show a parse error on CTRL+D/EOF anymore.
I've implemented this by adding a `NOP` variant to `REPLCmd` similar to how it was done in Idris 1 (as far as I understand :)). Is this the way to do this?


edwinb:
I think that's as good a way as any to do this. Thanks!

PR #81, opened 2019-08-13
Fix file handle leaks

### LINKED ISSUE #74: Interpreter does not clean opened files when loading/reloading files
The interpreter blows up very quickly when one loads a file too many times, yielding a `too many open files` error that seems irrecoverable.

# Steps to Reproduce

* Create some Idris source file `Source.idr`, possibly valid or invalid it does not matter
* Create an `input` file containing about 50 lines like:
   ```
    :l Source.idr
   ```
* Run `Idris2 < input`

# Observed Behavior

```
Text.Parse> 2/2: Building Text.Parse (Text/Parse.idr)
(implicit):1:1--1:1:Builtin not found
PrimIO> File error in Text/Parse.idr: Too many open files
Main> File error in Text/Parse.idr: Too many open files
Main> File error in Text/Parse.idr: Too many open files
Main> File error in Text/Parse.idr: Too many open files
Main> File error in Text/Parse.idr: Too many open files
Main> File error in Text/Parse.idr: Too many open files
Main> File error in Text/Parse.idr: Too many open files
Main> File error in Text/Parse.idr: Too many open files
```

# Expected Behavior

Files should be closed when reloaded and interpreter should not blow up unless one tries to load 100s or 1000s of files.

clayrat:
Fixes #74

### DIFF
--- a/src/Idris/ModTree.idr
+++ b/src/Idris/ModTree.idr
@@ -129,7 +129,8 @@ fnameModified fname
     = do Right f <- coreLift $ openFile fname Read
              | Left err => throw (FileErr fname err)
          Right t <- coreLift $ fileModifiedTime f
-             | Left err => throw (FileErr fname err)
+             | Left err => do coreLift $ closeFile f
+                              throw (FileErr fname err)
          coreLift $ closeFile f
          pure t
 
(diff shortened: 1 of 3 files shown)


edwinb:
Thanks!
This sort of thing is why I need a better type system :). I hope that QTT will help avoid this kind of problem, once we've learned how to use it properly...

PR #84, opened 2019-08-14
Tidy up REPL shutdown

akuhlens:
Previously, shutting down the REPL was hacky. Piping a file as input to the REPL or pressing Ctrl-D in the REPL caused a parse error. By adding an empty prompt case to the interpret function and modifying the end of file logic this PR keeps this error from happening.


edwinb:
This seems to have a conflict with your previous patch, which is probably my fault for being slow to merge. Would you mind fixing it up as you prefer, and then I'll merge? Thanks for the fix.

PR #89, opened 2019-08-20
Fixes #88: integerToNat hangs for negative integers

### LINKED ISSUE #88: `integerToNat` hangs for negative integers
# Steps to Reproduce

`integerToNat (2 - 3)`

# Expected Behavior

Either return `Z` or throw an error

# Observed Behavior

non terminating loop (well it will probably fail eventually when it gets to the last negative integer)

# Extra Info

I came across this because I am working on a project where we want to use natural numbers in both Haskell and Purescript. It turns out that Haskell natural numbers libraries all implement `Num` which in haskell has methods for `negate` and `-`. This means that the library authors chose a certain behaviour (crashing in most cases) in order to get `+` and `*`.

In purescript `+` and `*` are part of the `Semiring` class which is much nicer. There is an integerToNat function defined in `Data.Natural` but it's not part of any standard class because it makes a choice as to what should happen. Similarly there is also a `minus` function that is not part of any class for the same reason.

It seems to me that the way numeric classes work in Purescript is very well thought out and I've now come across a situation where that matters. Could we copy these numeric classes in Idris2 rather than the current classes (`Num`, `Neg` etc), they seem a little more `standard` in terms of it is clear what a `Semiring` is, there are agreed upon rules, the same can't be said for `Num`, for example it is quite different in Haskell and Idris.

So finally there should either be no `integerToNat` defined or it should either crash or return `Z` or possibly have 2 different versions that do different things. `Nat` should not implement a class that has `fromInteger` though.

christian-public:


edwinb:
Oh yes, good point :). Thanks!

PR #90, opened 2019-08-20
Added note to INSTALL.md to recommend Mac OS X users install Chez fro…

xoltar:
…m source since homebrew package appears to not support threading

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -24,7 +24,9 @@ Chez Scheme is available to build from source:
 
 + https://cisco.github.io/ChezScheme/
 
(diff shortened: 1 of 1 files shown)


edwinb:
I can't remember how I first installed it but I don't remember encountering this... anyway, it's good to have the note, thanks.

I'm actually thinking of switching to Racket as the default since it's easier to install, generates executables, and Racket on Chez generates code that's basically as fast as Chez.

edwinb:
I thought I'd merged this... apparently not! Sorry...

PR #96, opened 2019-08-29
Add version command

symbiont-arnaud-bailly:
This adds a `Version` (and corresponding `ShowVersion` in `REPL.idr`) command to the IDE Mode. It's a step towards making IDE Mode closer to Idris1 in order to make it possible to use Idris2 with emacs and idris-mode. I plan to implement more features in forthcoming PRs, at least to make starting REPL and loading file into it from idris-mode possible.

This also changes the way version is handled in the build (it's set as constants in `Makefile` and `YafflePaths.idr`) and introduces a `Idris.Version` module to expose it.


abailly:
@edwinb Is this PR potentially still useful or do you have other plans? I would be happy to work more on IDE integration but unsure how you want to deal with it.

edwinb:
@abailly I haven't really thought about this much, to be honest. I think this PR is nice, in any case. Thanks for fixing the conflicts!

I think it would still be nice to show the version in the startup banner. Perhaps we should have a --nobanner option like Idris 1 does, so that the tests don't need to be updated every time there''s an update. But that can be done independently of this patch.

ska80:
> Perhaps we should have a --nobanner option like Idris 1 does, so that the tests don't need to be updated every time there''s an update. But that can be done independently of this patch.

#102

abailly:
@ska80 That was fast!

PR #97, opened 2019-08-30
Fixed #!/usr/bin/env shebang for Chez backend

lodi:
On NixOS, idris2 can't find scheme in the usual locations, so it
defaults to generating the following shebang:

    #!/usr/bin/env scheme --script

The `env` program interprets `scheme --script` as one monolithic
command, instead of as a command and one argument.

    /usr/bin/env: ‘scheme --script’: No such file or directory
    /usr/bin/env: use -[v]S to pass options in shebang lines

The -S flag forces `env` to split on whitespace in the intuitive
manner.


edwinb:
Thanks!

PR #98, opened 2019-09-05
code golfing

clayrat:
Some minor rearrangements for readability I've made while familiarizing myself with the code base


edwinb:
These all seem fair enough. I need to teach my editor to stop adding spaces at the end of lines, or otherwise find a way to remove them automatically.

PR #99, opened 2019-09-13
remove another message to make network tests more reliable

abailly:
I noticed that there is still a race condition in the ordering of the messages between client and server, I _think_ this time it's pretty safe... (famous last words)


abailly:
I fixed the merge conflicts but tests failed. Pushed #99 to fix the tests.

edwinb:
Let's give it a go, thanks! Would this be better as part of the main test suite, though?

PR #101, opened 2019-09-20
reimpl: rigPlus & rigMult in Core/TT

luochen1990:
This PR is just a simple optmization of the code style, it is better at code length and runtime performance (since it has less branch so is expected to be more friendly to branch prediction).

### DIFF
--- a/src/Core/TT.idr
+++ b/src/Core/TT.idr
@@ -231,29 +231,23 @@ Show RigCount where
   show Rig1 = "Rig1"
   show RigW = "RigW"
 
+fromInt : Int -> RigCount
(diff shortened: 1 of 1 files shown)


edwinb:
Thanks for the patch. Do you have any data to show the performance improvement though? I don't see any difference when building the libraries, which isn't surprising, because these functions don't show up when profiling so I don't think they're a bottleneck at all. If there's any real gains to be had, they're mostly likely to be found in the evaluator (`Core.Normalise`)

I'm going to merge it anyway, because it is also a neater way to do this, and I would think less error prone than the existing lookup table.

PR #104, opened 2019-09-27
Add a fancy banner like in Idris 1

ska80:
I like the startup banner in Idris 1.  Can we have the same in Idris 2?


edwinb:
Ah yes. Why not :). Thanks!

PR #118, opened 2019-10-07
Parenthesize delayed types/terms in errors

clayrat:
Before this change, error messages involving delayed arguments would look a bit confusing:
```
Can't solve constraint between:
   maybe Delay foo Delay bar baz
...
```
Now they are printed as
```
maybe (Delay foo) (Delay bar) baz
```


edwinb:
Ah yes, this has been annoying me lately (but apparently not enough to do anything about it myself...) Thanks!

PR #119, opened 2019-10-08
Add some missing public exports

timsueberkrueb:
The following code is currently broken in Idris 2 but works in Idris 1:
```idris

bar : List Integer
bar = [0..9]

foo_works: Vect (length Main.bar) Integer
foo_works = fromList bar

foo_broken : Vect 10 Integer
foo_broken = fromList bar

works : length (the (List Nat) (1 :: 2 :: [])) = 2
works = Refl

doesNotWork : length Main.bar = 10
doesNotWork = Refl
```
Error message:
```
1/1: Building integerToNatProblem (integerToNatProblem.idr)
integerToNatProblem.idr:18:14--19:1:While processing right hand side of Main.foo_broken at integerToNatProblem.idr:18:1--19:1:
Can't solve constraint between:
	length bar
and
	S (assert_total (integerToNat (prim__sub_Integer 10 (fromInteger 1))))
```

Looks like this got stuck because `Vect.length` and `Prelude.countFrom` and `Prelude.takeUntil` (which are used in the `Prelude.rangeFromTo` implementation) were not `public export`.

I'm not sure whether `Prelude.countFrom` and `Prelude.takeUntil` are intended to be exposed though?

### DIFF
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -180,7 +180,7 @@ merge = mergeBy compare
 ||| ```idris example
 ||| reverse [1,2,3,4]
 ||| ```
-export
+public export
 reverse : Vect len elem -> Vect len elem
 reverse xs = go [] xs
   where go : Vect n elem -> Vect m elem -> Vect (n+m) elem
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -17,7 +17,7 @@ data Vect : (len : Nat) -> (elem : Type) -> Type where
 -- Hints for interactive editing
 %name Vect xs,ys,zs,ws
 
-export
+public export
 length : (xs : Vect len elem) -> Nat
 length [] = 0
 length (x::xs) = 1 + length xs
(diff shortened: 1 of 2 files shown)


timsueberkrueb:
`Vect.reverse` (used by `Vect.fromList`) seems to be another one.

Example case:

```
Can't solve constraint between:
	0 :: tail (reverse [9, 8, 7, 6, 5, 4, 3, 2, 1, 0])
and
	reverse (fromList' [] (rangeFromTo (fromInteger 0) (fromInteger 9)))
```

```idris
foo : Vect 10 Integer
foo = fromList [0..9]

bar : Dec (UniqElem 0 Tests.foo)
bar = isUniqElem 0 foo

bar0IsUniqElem : {auto notInTail: NotIn 0 (tail Tests.foo)}
        -> UniqElem 0 Tests.foo
bar0IsUniqElem {notInTail} = Here notInTail
```

edwinb:
Ah yes, I'm encountering this sort of thing from time to time. I've erred on the side of not putting `public` in but it really needs to be there for things which are intended for use in proofs. I'm not so sure about reverse though.

edwinb:
I'm not sure `countFrom` and `takeUntil` should be exposed either, come to think of it, but since they are part of how the range syntax works, and we want that computing in types, I suppose they'll have to be. This probably needs more thought, but for the sake of making things work as in Idris 1 I'll merge this and come back to the problem later.

PR #123, opened 2019-10-15
allow matching constants in with clauses 

### LINKED ISSUE #122: With clauses fail to match constants
If possible, please include complete self contained source files (for example by uploading to https://gist.github.com) that exhibit the issue in addition to quoting from them here. The smaller the example the better!

# Steps to Reproduce

Create `Temp.idr` with the following code:

```
module Temp

import Data.List

foo : (l : List Int) -> String -> Int
foo game "length" with (length l)
  foo game "length" | 0 = "Zero"
  foo game "length" | n = "Not Zero"
foo game s = "Other"
```

then 

```
idris2 --no-banner Temp.idr
```

# Expected Behavior

```
1/1: Building Temp (Temp.idr)
```

# Observed Behavior

```
1/1: Building Temp (Temp.idr)
Temp.idr:6:7--6:16:With clause does not match parent
```

abailly:
fix #122

### DIFF
--- a/src/TTImp/WithClause.idr
+++ b/src/TTImp/WithClause.idr
@@ -19,7 +19,7 @@ mutual
       = if n == n' && isSuffixOf ns' ns then pure [] else matchFail loc
   getMatch lhs (IVar _ (NS ns n)) (IVar loc n')
       = if n == n' then pure [] else matchFail loc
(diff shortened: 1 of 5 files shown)


edwinb:
Thanks! I should have remembered to test this case at the time, given that it's a mistake I've made before...

PR #124, opened 2019-10-16
Fix --clean option

clayrat:
This fixes two bugs:
* Build artefacts are in `ttc` subfolder as of https://github.com/edwinb/Idris2/commit/65db4fbf961cbdd36f10a9a75e7c3311b91f5585
* Namespaces were reversed at the wrong place leading to problems with nested folders

Also adds a few `Show` instances (mostly for debugging)

### DIFF
--- a/src/Data/StringMap.idr
+++ b/src/Data/StringMap.idr
@@ -11,16 +11,16 @@ Key = String
 
 private
 data Tree : Nat -> Type -> Type where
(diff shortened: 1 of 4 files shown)


edwinb:
Thanks! The intention with the `ttc` subfolder is that there could be other subfolders. Most obviously, one for executables (which will also need cleaning at some point) but I haven't got around to that yet.

PR #125, opened 2019-10-16
Add `head` and `tail` to base/Data.List.

msmorgan:
These functions use the NonEmpty predicate type in order to prove that the operation will be valid.

Implementations copied from Idris1's Prelude.List module, except without expanding the auto implicit argument.


edwinb:
Good idea, thanks

Issue #133, opened 2019-10-22
Idris2 stackoverflows


ohad:
# Steps to Reproduce
Download the attached file [stackoverflow.zip](https://github.com/edwinb/Idris2/files/3759034/stackoverflow.zip) unzip, unlimit process stack size, and compile `CMonoids.idr`:

```
$ unzip stackoverflow.zip
Archive:  stackoverflow.zip
  inflating: PreorderReasoning.idr  
  inflating: VectReasoning.idr  
  inflating: Carriers.idr      
  inflating: Signatures.idr    
  inflating: Algebras.idr      
  inflating: Presentations.idr  
  inflating: Models.idr        
  inflating: Frex.idr          
  inflating: Monoids.idr       
  inflating: CMonoids/Axioms.idr  
  inflating: Powers.idr        
  inflating: CMonoids/Notation.idr  
  inflating: CMonoids/Nat.idr  
  inflating: CMonoids/Semiring.idr  
  inflating: CMonoids/Free.idr  
  inflating: CMonoids.idr
$ ulimit -s unlimited      
$ idris2 -c CMonoids.idr 
```
# Expected Behavior
```
1/16: Building PreorderReasoning (PreorderReasoning.idr)
2/16: Building VectReasoning (VectReasoning.idr)
3/16: Building Carriers (Carriers.idr)
4/16: Building Signatures (Signatures.idr)
5/16: Building Algebras (Algebras.idr)
6/16: Building Presentations (Presentations.idr)
7/16: Building Models (Models.idr)
8/16: Building Frex (Frex.idr)
9/16: Building Monoids (Monoids.idr)
10/16: Building CMonoids.Axioms (CMonoids/Axioms.idr)
11/16: Building Powers (Powers.idr)
12/16: Building CMonoids.Notation (CMonoids/Notation.idr)
13/16: Building CMonoids.Nat (CMonoids/Nat.idr)
14/16: Building CMonoids.Semiring (CMonoids/Semiring.idr)
15/16: Building CMonoids.Free (CMonoids/Free.idr)
16/16: Building CMonoids (CMonoids.idr)
$
```
# Observed Behavior
```
1/16: Building PreorderReasoning (PreorderReasoning.idr)
2/16: Building VectReasoning (VectReasoning.idr)
3/16: Building Carriers (Carriers.idr)
4/16: Building Signatures (Signatures.idr)
5/16: Building Algebras (Algebras.idr)
6/16: Building Presentations (Presentations.idr)
[...]

PR #135, opened 2019-10-24
Bugfix #133

ohad:
turn `HasNames (List a)` tail-recursive 

The problem seems to be a non-tail-recursive traversal of the huge proof terms I'm constructing.
Specifically the `HasNames` implementation for `List a` in `Core/Binary.idr` was not tail recursive.
Making it tail recursive makes the stack overflow go away.

### DIFF
--- a/src/Core/Binary.idr
+++ b/src/Core/Binary.idr
@@ -62,12 +62,16 @@ record TTCFile extra where
   extraData : extra
 
 HasNames a => HasNames (List a) where
-  full c [] = pure []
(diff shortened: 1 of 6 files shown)


ohad:
@edwinb if you can confirm the order of names in the result of `full`/`resolved` doesn't matter, then I'll drop the `reverse` calls, which might end up in a little speed-up.

edwinb:
Ah yes, a stack overflow usually indicates something like that. I wonder if there are other things like this lurking. Thanks for the fix.

PR #136, opened 2019-10-24
Add 'sourcedir' option to IPKG

clayrat:
Useful for getting more readable diffs between Idris2 and Idris1 branches of code :)


edwinb:
Thanks, I've been wanting to do this but not got around to it...

PR #137, opened 2019-10-25
Remove trailing whitespace from Idris sources.

msmorgan:
This is the result of running the command:

$ find . -name '*.idr' -type f -exec sed -i -E 's/\s+$//' {} +

I confirmed before running it that this would not affect any markdown
formatting in documentation comments.

### DIFF
--- a/libs/base/Control/Monad/Identity.idr
+++ b/libs/base/Control/Monad/Identity.idr
@@ -1,6 +1,6 @@
 module Control.Monad.Identity
 
-public export 
+public export
 record Identity (a : Type) where
   constructor Id
   runIdentity : a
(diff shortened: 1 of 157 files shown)


edwinb:
Thanks, I need to teach my editor to stop doing this or it'll just happen again...

buzden:
> Thanks, I need to teach my editor to stop doing this or it'll just happen again...

You can at least teach your vim to *show* you trailing whitespaces by adding something like
```
highlight ExtraWhitespace ctermbg=red guibg=red
match ExtraWhitespace /\s\+\%#\@<!$/
```
to your `~/.vimrc` file.

PR #143, opened 2019-10-27
Bugfix #133 and library support for tail-recursive list manipulation

ohad:
Fix a stack overflow with a tail recursive list integer length function .

This PR replaces #142, but if you don't want to merge in #142, I can make a separate one with the bugfix only. Here's the description for the (retracted) #142 :
 
I've made tail-recursive versions of the `Data.List` functions, as Idris has been stack-overflowing on Frex with lots of stack frames from `List.length`.

Since we might want to also reason about the tail-recursive versions inductively, I've also included proofs they are all extensionally equivalent to their non-tail recursive counterparts.

I mostly just bashed this out, so lots of rooms to refactor it, and if there's something specific that bothers you, just tell me and I'll change it.


edwinb:
Thanks! Just checking my merge didn't break anything...

PR #145, opened 2019-10-29
allow overriding racket/chicken paths through environment variables

lodi:
This is to help Idris2 generate correct scheme output on NixOS where
Racket and Chicken aren't installed in the standard locations, and
the /usr/bin/env trampoline is disabled at package build time.
The Chez backend already supports overrides.

Also, fixed the test runner to restore the correct working directory
after a failed test, and fixed the top-level Makefile to allow the
IDRIS2_VERSION variable to be queried without building the project.

### DIFF
--- a/tests/Main.idr
+++ b/tests/Main.idr
@@ -82,25 +82,31 @@ fail err
 runTest : String -> String -> IO Bool
 runTest prog testPath
     = do chdir testPath
-         putStr $ testPath ++ ": "
-         system $ "sh ./run " ++ prog ++ " | tr -d '\\r' > output"
-         Right out <- readFile "output"
-               | Left err => do print err
-                                pure False
-         Right exp <- readFile "expected"
-               | Left err => do print err
-                                pure False
-
-         if (out == exp)
-            then putStrLn "success"
-            else do
-              putStrLn "FAILURE"
-              putStrLn "Expected:"
-              printLn exp
-              putStrLn "Given:"
-              printLn out
+         isSuccess <- runTest'
          chdir "../.."
-         pure (out == exp)
+         pure isSuccess
+    where
+        runTest' : IO Bool
+        runTest'
+            = do putStr $ testPath ++ ": "
+                 system $ "sh ./run " ++ prog ++ " | tr -d '\\r' > output"
+                 Right out <- readFile "output"
+                     | Left err => do print err
+                                      pure False
+                 Right exp <- readFile "expected"
+                     | Left err => do print err
+                                      pure False
+
+                 if (out == exp)
+                    then putStrLn "success"
+                    else do
+                      putStrLn "FAILURE"
+                      putStrLn "Expected:"
+                      printLn exp
+                      putStrLn "Given:"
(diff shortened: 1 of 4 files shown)


tests/Main.idr
```diff
@@ -82,25 +82,31 @@ fail err
...
-
-         if (out == exp)
-            then putStrLn "success"
-            else do
-              putStrLn "FAILURE"
-              putStrLn "Expected:"
-              printLn exp
-              putStrLn "Given:"
-              printLn out
+         isSuccess <- runTest'
```
msmorgan, line 85: This function doesn't appear to do anything different than the unmodified version. Can you explain this change?

lodi: The old code was forgetting to "cd ../.." after a failing test, so I separated the working directory push/pop logic from the actual test harness.  The harness code is unmodified, just indented a few more stops.

I discovered this trying to build on NixOS where half the tests appeared to be failing but actually it was just one of the tests in the middle that had a shebang issue.

msmorgan: Wouldn't the updated function have the same behavior as the original? If `runTest'` causes an IO error (which is likely what happened with the shebang issue), chdir still won't run.

lodi: I think the diff view is obscuring the structure of the updated function:

    runTest : String -> String -> IO Bool
    runTest prog testPath
        = do chdir testPath
             isSuccess <- runTest'
             chdir "../.."
             pure isSuccess
        where
            runTest' : IO Bool
            runTest'
                = do putStr $ testPath ++ ": "
                     system $ "sh ./run " ++ prog ++ " | tr -d '\\r' > output"
                     Right out <- readFile "output"
                         | Left err => do print err
                                          pure False
                     Right exp <- readFile "expected"
                         | Left err => do print err
                                          pure False

                     if (out == exp)
                        then putStrLn "success"
                        else do
                          putStrLn "FAILURE"
                          putStrLn "Expected:"
                          printLn exp
                          putStrLn "Given:"
                          printLn out
    
                     pure (out == exp)

Whether or not `runTest'` succeeds or fails, chdir will be called on the way out of `runTest`.  I checked that this does recover after a failing test.

msmorgan: That was never a problem to begin with, though, if a test fails naturally. I tried it myself, by changing the expected output of `idirs2/basic001`. The runner reports the test as a failure and continues running the rest of the tests without a hitch.

When a test fails because of an IO error, however, both versions of this function will not change directory. They will exit early from the `do` block because of the error.

msmorgan: Oh, I see what you mean now. Somehow I missed the guard assignment for readFile. So your change fixes the situation when the output or expected file can't be read.

lodi: Yes, exactly.  I ran into this in `tests/ideMode/ideMode002` where `gen_expected.sh` with an unpatched shebang would fail to generated `expected`, which then causes an early exit from `runTest`.

edwinb:
Thanks!

PR #147, opened 2019-10-29
Add zip functions to base/Data.List.

msmorgan:
This includes the following functions: zipWith, zip, zipWith3, zip3


edwinb:
Thanks for this! So far I've been adding things as I need them, but we'll really need to think about libraries properly at some stage.

PR #153, opened 2019-11-11
export Network.Socket functions

abailly:


libs/network/Network/Socket.idr
```diff
@@ -152,7 +160,7 @@ recv sock len = do
 ||| Returns on success the payload `String`
 |||
 ||| @sock The socket on which to receive the message.
-partial
+export
```
sysvinit, line 163: This should probably be `export partial` instead of just `export` to preserve the existing totality modifier.

edwinb: Probably, and it's likely to cause a compile failure when I make progress on the totality checker, but I'll still merge as it is.

PR #156, opened 2019-11-12
Misc fixes and additions

clayrat:
Some minor rearrangments in the compiler source + add some missing library functions from Idris1 needed for the port of https://github.com/gallais/idris-tparsec


edwinb:
Thanks for this! I did consider `Alternative` for the Prelude but my feeling was that it's not quite satisfying the criterion of "used by almost any non-trivial program". No doubt different people will have different opinions about this, but for now could it be in `base` instead please?

clayrat:
But `Alternative` is already in the Prelude, this only adds `choice` and `choiceMap` for it. Would you like to have just these two functions moved to `base`?

edwinb:
Well that's odd, I've obviously changed my mind too many times :). I'll just merge...

PR #158, opened 2019-11-15
Add List.replicate from Idris 1

ska80:


edwinb:
Looks like another PR beat you to it! So, thanks for the addition, but I should probably close this.

PR #164, opened 2019-11-19
Prelude linearity

ohad:
(Based on PR #143 , if you want one based on HEAD, let me know.)

Add linearity annotations to the prelude ( 65c151f  ).
Since that affects line numbers, also update one test.

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -85,7 +85,7 @@ span p (x::xs) =
 
 public export
 break : (a -> Bool) -> List a -> (List a, List a)
-break p = span (not . p)
+break p xs = span (not . p) xs
 
 public export
 split : (a -> Bool) -> List a -> List (List a)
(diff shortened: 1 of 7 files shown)


edwinb:
This is good to get right, thanks! But there seems to be a lot of other unrelated things here... In particular, I would like some kind of preorder reasoning syntax, but I don't yet know the best way to do it so I don't think it should be in `base` yet. Perhaps making a `contrib` library and adding there would be better for the moment?

ohad:
Sorry, I only noticed this discussion now (oops!).

> In particular, I would like some kind of preorder reasoning syntax, but I don't yet 
> know the best way to do it so I don't think it should be in base yet

Yes, I was a bit sloppy with this pull request and based it on the previous PR #143, which has been merged into master now.

This means the preorder stuff is now on `base` on `master`, which I guess you'd rather not have?
Shall I sort this out in 2 pull requests?
1. move `PreorderReasoning` from `base` to `conrib`
2. a rebase of the linearity annotations patch on master.

edwinb:
Yes, that sounds like a good way around. Although we don't have a `contrib` yet, but this is probably as good a place as any to start.

ohad:
Done in PRs #180 , #181 , making this PR irrelevant.

PR #165, opened 2019-11-20
Remove Vim swap files

ska80:


edwinb:
Well this is strange, but clearly they shouldn't be there! Thanks.

PR #167, opened 2019-11-25
Add unwords and unlines to base/Data.Strings from Idris 1

ska80:


edwinb:
Good to have, thanks!

PR #172, opened 2019-12-18
Use HTTPS instead of HTTP

MilanKral:


edwinb:
Seems reasonable, thanks

PR #173, opened 2019-12-21
Reorder Scheme compilers order to pick generic Scheme last.

timmyjose:
This fixes the issue where some Scheme compiler like mit-scheme is
installed and that gets picked up first, leading to an infinite
loop in the tests.

Also see attached image in the P.R:

On a system with `mit-scheme` as the default Scheme compiler, the following step hangs forever as part of `make install`:

Screenshot 2019-12-21 at 6 13 40 PM

My  understanding is that the `mit-scheme` instance simple starts a REPL that will wait around forever for input that is never received.


edwinb:
Thanks for the fix!

PR #176, opened 2020-01-01
Irrelevance stdlib

ohad:
Add missing irrelevance annotations to standard library ( 43ed2f9 )

Data.Vect.Elem
Hetrogeneous and Homogeneous equality

Also (minor): fix typo in a comment ( e6debc9 )

(seemed silly to put this in a separate pull request)

### DIFF
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -808,9 +808,9 @@ implementation {n : Nat} -> Traversable (Vect n) where
 
 ||| A proof that some element is found in a vector
 public export
-data Elem : a -> Vect k a -> Type where
-     Here : Elem x (x::xs)
-     There : (later : Elem x xs) -> Elem x (y::xs)
+data Elem : (0 x : a) -> (0 xs : Vect k a) -> Type where
+     Here : {0 x : a} -> Elem x (x::xs)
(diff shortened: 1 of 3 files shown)


edwinb:
Are these necessary? Do they make something work that otherwise doesn't? If so, something might be wrong... for `Elem` the annotations are unnecessary because they are implicits and thus default to 0, and for `===` they're in types so irrelevant anyway.

ohad:
They make the following work:
```
foo : {0 x : a} -> Type
foo {x} = x ~=~ x
```
With `head` on master I get:
```
While processing right hand side of Main.foo at Foo.idr:5:1--6:1:
x is not accessible in this context
```
whereas with `irrelevance-stdlib` it type-checks fine.

Should I open a bug report then?

edwinb:
A better way to do this would be to say that `foo` is meant to be run in irrelevant context, I think. All this behavious is expected.

ziman:
I agree with Edwin that `0 foo : {0 x : a} -> Type` is the right way to approach this.

edwinb:
I've left this open for a while, but I think the conclusion is that the right solution is to use `0` multiplicity functions, so I'll close now.

ohad:
I'm fine with it, have adopted this way of doing it. Thanks!

PR #177, opened 2020-01-01
Fix chez010 test on MacOS

SPY:
Hi

After removing explicit referencing of libcb.so in [last commit](https://github.com/edwinb/Idris2/commit/bf5b2298e068ebeaddd70b695426f52ebe67c1ab) I've got a test failure on MacOS:
```
Exception: (while loading libcb.dylib) dlopen(libcb.dylib, 2): image not found
chez/chez010: FAILURE
```

Chez backend uses different extensions for dynamic libraries depending on OS, so in implicit mode it looks for libcb.dylib on macos, but the test provides libcb.so only.
I solved this problem by generating both libs, but maybe test runner should be smarter?

Best,
Ilya


edwinb:
Possibly instead the back end should be cleverer about how it finds libraries. I don't know a good way of doing this portably, so any ideas are welcome.

Possibly a better temporary hack would be to change `CB.idr` in this test to refer to `libcb.so`, which it used to do before expressions were allowed in `%foreign` directives. If that works, I think I'd prefer it to this fix.

SPY:
> Possibly instead the back end should be cleverer about how it finds libraries. I don't know a good way of doing this portably, so any ideas are welcome.

I agree with it. It is pretty common to have *.so files on Mac OS, especially for apps installed from brew.

> Possibly a better temporary hack would be to change CB.idr in this test to refer to libcb.so, which it used to do before expressions were allowed in %foreign directives. If that works, I think I'd prefer it to this fix.

I did it this way(implicit extension) and it works for now.

edwinb:
Thanks!

PR #180, opened 2020-01-09
Create a contrib directory in the standard library

ohad:
Move the TailRec.idr and current poor-man's syntax for:
  equational reasoning (`Syntax.PreorderReasoning`)
  with proof           (`Syntax.WithProof`)

(See discussion in PR #164 .)


edwinb:
Thanks! I've been meaning to have something like this for a while, so good to make a start.

PR #181, opened 2020-01-09
Linearity annotation in std-lib

ohad:
Add linearity annotations to the prelude.
Since that affects line numbers, also update one test.

### DIFF
--- a/libs/prelude/Prelude.idr
+++ b/libs/prelude/Prelude.idr
@@ -67,7 +67,8 @@ the : (0 a : Type) -> (1 x : a) -> a
 the _ x = x
 
 public export %inline
-id : (x : a) -> a
+id : (1 x : a) -> a           -- Hopefully linearity annotation won't
+                              -- break equality proofs involving id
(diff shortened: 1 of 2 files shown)


edwinb:
Looks like I've messed up the line numbers with some other merges, but I think it is good to have these annotations accurate when we can. Thanks for that!

LeifW:
Now that Nat, Int, and Integer have their Num.+ impls marked as linear - I'm wondering if we'd want Num.+ itself marked as linear? It seems for most kinds of things we'd call numbers, we'd want + to use both args. But maybe we want to leave it flexible?

ohad:
I'm wary of restricting the _interface_ to be left-linear.
Do you have an example of which client code could rely on `Num.+` instances being left-liner?

edwinb:
I'm a bit scared of changing the interface too, as things are. There are other possible ways of implementing plus even for Nat that might be linear in the right argument. I think we need to get a bit more experience with this...

PR #182, opened 2020-01-09
Tidying up Inline.idr for streaming

andrevidela:
This is a demo PR made on stream. It only contains aesthetic changes


andrevidela:
this PR shouldn't break anything but the tests don't pass on my machine on test `chez/chez010`

```
Exception: (while loading libcb.dylib) dlopen(libcb.dylib, 2): image not found
```

I am missing something?

edwinb:
It should be fine since #177...

PR #185, opened 2020-01-15
MonadError, the errorful Monad transformer

alrunner4:
Utilities for working simply with mainly IO-threaded programs.

### DIFF
--- a/libs/base/Control/Monad/Error/IORef.idr
+++ b/libs/base/Control/Monad/Error/IORef.idr
@@ -1,0 +1,11 @@
+module Control.Monad.Error.IORef
+import Control.Monad.Error
+import Data.IORef
+
+public export
+overIORef : IORef a -> ( a -> IOError e a ) -> IOError e a
+overIORef ref xform = do
+    val <- lift( readIORef ref )
+    newval <- xform val
+    lift( writeIORef ref newval )
+    pure newval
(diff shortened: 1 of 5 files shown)


libs/base/Control/Monad/Error/IORef.idr
```diff
@@ -0,0 +1,11 @@
+module Control.Monad.Error.IORef
+import Control.Monad.Error
+import Data.IORef
+
+public export
+overIORef : IORef a -> ( a -> IOError e a ) -> IOError e a
+overIORef ref xform = do
+    val <- lift( readIORef ref )
```
alrunner4, line 8: Occurred to me: _should_ `lift` be in scope, despite no `import Control.Monad.Trans` here, and no `import public` in `Control.Monad.Error`? I'm guessing that it's visible because it's part of an interface...?

edwinb:
There's a test failure here due to the prelude change (I should do something about the line number display there, it is fairly useless.)

Still, I'm not sure this should be in `base` - it's probably better in `contrib` for now, because I would rather not encourage monad transformers for this sort of thing. Of course people might want to do things that way, which is fine, and it's understandable if people want to do things in a way which is familiar from Haskell, but I'm shortly going to add an error+state+IO type which I've been finding works quite nicely.

(I know we have State and Trans in base too. Should we?)

PR #189, opened 2020-01-31
Fix IDE Mode for Atom

christian-public:
Unfortunately a change in 432cc7c breaks the `type-of` (and maybe other commands) in the Atom IDE plugin. By looking at the output of the IDE Mode before this change I could see that there was a missing argument in the return value (See example at the bottom). The extra argument is used for highlighting, though it is currently not in use, as noted by the previous implementation of `printWithStatus`:
https://github.com/edwinb/Idris2/blob/a145ba99d7bda290c1b0ba189849d5b764795bba/src/Idris/REPLCommon.idr#L36-L39

@abailly Were there any specific reasons for removing this extra argument in the IDE Mode?

I am sorry if this pull request breaks any other IDE integrations (though the format should now be more similar to Idris 1's format).

---

IDE Mode before 432cc7c (Note the extra `()` argument.):
```
000015((:type-of "foo") 1)
000049(:return (:ok "-------------------------------------
foo : IO ()" ()) 1)
```

IDE Mode after 432cc7c:
```
000015((:type-of "foo") 1)
000046(:return (:ok "-------------------------------------
foo : IO ()") 1)
```

EDIT: Forgot to fix the tests. Fixed now.

### DIFF
--- a/src/Idris/IDEMode/REPL.idr
+++ b/src/Idris/IDEMode/REPL.idr
@@ -179,7 +179,9 @@ idePutStrLn outf i msg
 
 printIDEWithStatus : File -> Integer -> String -> SExp -> Core ()
 printIDEWithStatus outf i status msg
-    = do let m = SExpList [SymbolAtom status, toSExp msg ]
+    = do let m = SExpList [SymbolAtom status, toSExp msg,
+                           -- highlighting; currently blank
+                           SExpList []]
          send outf (SExpList [SymbolAtom "return", m, toSExp i])
 
 printIDEResult : File -> Integer -> SExp -> Core ()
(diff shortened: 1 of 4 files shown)


abailly:
@chrrasmussen I did that a while ago. I removed this because Emacs' idris-mode went happily without it and we did not have a good story about syntax highlighting at that time. I don't think this will break anything but I can give it a try on Emacs if you want.

christian-public:
@abailly I imagined it was something like that :-) It would be great if you could check. Thank you!

edwinb:
I've merged to get the Atom mode working again, though we really need to get the highlighting working again for Emacs.

abailly:
I have started working again on the highlighting in Emacs and this change, as expected, breaks emacs-mode: The `:ok` token is expected to have a single value and we give it 2. Given this second argument is unused, would it be possible to change the Atom plugin instead?

abailly:
To be more precise: Highlighting information is sent in IDEMode "on the fly", as additional `:output` messages sent before the final `:return` message.

PR #195, opened 2020-02-23
Add support for `%default [total | covering | partial]` directive

ohad:
Design decisions:

0. This doesn't actually add any functionality as the totality checker is still a stub.

1. default totality annotations are session-local, so should be reset
   between files. So I added another piece of state to the `ElabDirectives` 
   under the assumption that it resets for each file.

2. the default default is `PartialOK`.

3. we won't annotate a definition with `PartialOK` unless the user
   inserts one themselves, relying on the previous point (2).

4. Internally, we reuse the `TotalReq` type in both the syntax tree (`PClaim` etc.) and the `TT` constructs. This simplifies one function slightly.

### DIFF
--- a/src/Core/Context.idr
+++ b/src/Core/Context.idr
@@ -123,9 +123,6 @@ data Clause : Type where
                 (lhs : Term vars) -> (rhs : Term vars) -> Clause
 
 public export
-data TotalReq = Total | CoveringOnly | PartialOK
-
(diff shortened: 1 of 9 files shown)


edwinb:
Cheers. I've also just changed 'default' to be a keyword to support default implicits, so I'll update this accordingly...

PR #197, opened 2020-02-23
Bring `src/Text/Parser` and its dependencies closer to idris2

ohad:
* Removing `%access` directives and inlining their visibility modifiers
* Fixing irrelevance annotations
* Including some standard libraries

The result is still not idris2 ready:

1. Incompatibility with the new Laziness mechanism, as it's missing
   some previous laziness/codata bindings

2. Some standard library functions moved around, and are now found in
   `Data.Nat`, that doesn't exist in Idris1

3. Annotating the `Functor (Grammar ... c ...)` implementation with a 
   relevance annotation for `c`, which seems to be incompatible with Idris1

### DIFF
--- a/src/Control/Delayed.idr
+++ b/src/Control/Delayed.idr
@@ -21,6 +21,7 @@ lazy = delayed LazyValue
 
 ||| Conditionally delay a value.
 export
-delay : (d : Bool) -> a -> delayed r d a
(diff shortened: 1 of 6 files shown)


edwinb:
Thanks - I definitely prefer not having the `%access` but this was mostly copied from the Idris 1 libraries...

PR #199, opened 2020-02-26
Improve 'idris2c' makefile target

ska80:


edwinb:
That does look better, thanks. This is step 0 towards making a distribution that doesn't require having an Idris 1 available.

PR #202, opened 2020-02-27
Dependent record example "length" field naming

glmxndr:


edwinb:
Oops, yes, this is the sort of thing that happens when you edit the text after pasting in working code! Thanks!

PR #203, opened 2020-02-27
Add PATH lookup to find Chez Scheme

m-bd:
On some systems "/usr/bin" and "/usr/local/bin" are not used.
If CHEZ env variable is not defined, lookup for chez in the PATH if
defined. Otherwise falls back to the old behaviour. (ie. lookup in
"/usr/bin" and "/usr/local/bin" as a last resort)

### DIFF
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -19,14 +19,20 @@ import System.Info
 
 %default covering
 
+
+pathLookup : IO String
+pathLookup
+    = do path <- getEnv "PATH"
+         let pathList = split (== ':') $ fromMaybe "/usr/bin:/usr/local/bin" path
+         let candidates = [p ++ "/" ++ x | p <- pathList,
+                                           x <- ["chez", "chezscheme9.5", "scheme"]]
+         e <- firstExists candidates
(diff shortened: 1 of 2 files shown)


anqur:
Well, just asking, will `/usr/bin/env chez` be better here?

m-bd:
On my different systems defaulting to `env scheme` similarly to racket and chicken without looking in `/usr/{,local/}bin` would be the best. (NixOS and nix/ubuntu)
But this may not work for some other people.

I initially encountered an issue with `lib-test` hanging forever while building idris2. It picked `mit-scheme` (installed globally) despite being in a pure nix env.

Hard coded paths are evil.

jbsolomon:
> 
> 
> Well, just asking, will `/usr/bin/env chez` be better here?

Please don't make MinGW/MSYS2/Cygwin mandatory on Windows.  Windows only supports PATH natively; a portable solution should use that.

edwinb:
This does seem like a better approach than the current one that just guesses. Thanks!

There are probably lots of things that don't work right on Windows. I don't have easy access to a Windows machine, but I would like to support it properly (I don't even really know what "properly" means here).

jbsolomon:
@edwinb if there's anything I can help take a look at on Windows, I'd be glad to!

PR #205, opened 2020-02-29
first stab at syntax highlighting in Emacs

abailly:
got the basic infrastructure to output syntax highlight in IDE mode set up but the results are far from being usable, need more work to understand how to reap information from Metadata

@edwinb Could you tell me which data structures I should look into in order to get the information I need? 

@chrrasmussen I have had to revert the change you made to add empty list in IDE mode's result output because it breaks emacs' idris mode. Perhaps we could modify the Atom plugin instead?


abailly:
Here is what I have so far...

Capture d’écran 2020-02-29 à 13 53 15

christian-public:
That's a bummer.

It looks like `:ok` is sent with either 1 or 2 parameters depending on context/message. Starting a new session with Idris 1's IDE mode gives the following output:
```
000018(:protocol-version 1 0)
00002d(:write-string "Type checking ./Main.idr" 0)
00020f(:output (:ok (:highlight-source ((((:filename "Main.idr") (:start 1 8) (:end 1 9)) ((:name "IO") (:implicit :False) (:key "AAAAAAAAAAACSU8=") (:decor :function) (:doc-overview "Interactive programs, describing I/O actions and
returning a value.") (:type "Type -> Type"))) (((:filename "Main.idr") (:start 1 11) (:end 1 12)) ((:name "Unit") (:implicit :False) (:key "AAAAAAAAAAAEVW5pdA==") (:decor :type) (:doc-overview "The canonical single-element type, also known as
the trivially true proposition.") (:type "Type")))))) 0)
000108(:output (:ok (:highlight-source ((((:filename "Main.idr") (:start 1 1) (:end 1 4)) ((:name "Main.main") (:implicit :False) (:key "AQAAAAAAAAAABG1haW4AAAAAAAAAAQAAAAAAAAAETWFpbg==") (:decor :function) (:doc-overview "") (:type "IO ()") (:namespace "Main")))))) 0)
000108(:output (:ok (:highlight-source ((((:filename "Main.idr") (:start 2 1) (:end 2 4)) ((:name "Main.main") (:implicit :False) (:key "AQAAAAAAAAAABG1haW4AAAAAAAAAAQAAAAAAAAAETWFpbg==") (:decor :function) (:doc-overview "") (:type "IO ()") (:namespace "Main")))))) 0)
0000e0(:output (:ok (:highlight-source ((((:filename "Main.idr") (:start 2 8) (:end 2 11)) ((:name "Main.foo") (:implicit :False) (:key "AQAAAAAAAAAAA2ZvbwAAAAAAAAABAAAAAAAAAARNYWlu") (:decor :metavar) (:namespace "Main")))))) 0)
000018(:set-prompt "*Main" 0)
000015(:return (:ok ()) 0)
000015((:type-of "foo") 1)
00025b(:return (:ok "--------------------------------------
foo : IO ()" ((39 3 ((:name "foo") (:implicit :False) (:key "AAAAAAAAAAADZm9v") (:decor :metavar) (:type "IO' (MkFFI C_Types String String) ()"))) (45 2 ((:name "IO") (:implicit :False) (:key "AAAAAAAAAAACSU8=") (:decor :function) (:doc-overview "Interactive programs, describing I/O actions and
returning a value.") (:type "Type -> Type"))) (48 2 ((:name "Unit") (:implicit :False) (:key "AAAAAAAAAAAEVW5pdA==") (:decor :type) (:doc-overview "The canonical single-element type, also known as
the trivially true proposition.") (:type "Type"))))) 1)
```

From the above example:
- `:output` has an `:ok` with a single parameter
- `:return` has an `:ok` with a single parameter as a response to `:set-prompt`
- `:return` has an `:ok` with two parameters as a response to `:type-of`

Do you think we could change Idris 2 IDE mode to accomodate this? Hopefully this would make it the IDE mode work in both/all editors. I can help to track down which messages returns what.

abailly:
Yes, I think it makes sense. I will change `Idris/IDEMode/REPL.idr` accordingly, it's pretty easy (and I will add a test).

edwinb:
@abailly which information is it you need? To get the type of a name, for example? I think what you have is the way I'd do it too, as things stand. But perhaps there could be better ways.

I haven't thought much about the IDE mode and highlighting, but I think there are two considerations: 

* Firstly, highlighting the source file. This is something the elaborator could keep track of and store directly in the metadata, since it has all the information.
* Secondly, highlighting outputs of other REPL commands (e.g. evaluation). This is rather harder, and probably needs a better pretty printer to keep track of location information in the output. It seems that for this we could be making the pretty printer's job easier by holding more information in the PTerm structure about the name type of a PRef.

Meanwhile, is this safe to merge now? How is the conflict with Atom?

abailly:
@edwinb no, it's not ready to merge, I still need to make the change as discussed with @chrrasmussen , will do it today and ping you when it's ready. 

re Syntax highlighting: OK, I understand that we don't have the information in `Metadata` yet, so we need to modify the elaboration process to add this metadata while it's processing the source file. Is that correct? That's interesting, it will give me the opportunity to do a deep dive in the elaboration process! What's a good entry point to start looking at it?

edwinb:
`TTImp.Elab.Term` is where TTImp terms get elaborated to TT, via a top level function `checkTerm`. Probably the specific place to look for name highlighting is `TTImp.Elab.App`. This is probably the most complicated bit of the whole process since it needs to deal with implicit arguments (that's why variables are considered applications - they might turn out to be implicit function applications after all).

If the plan is to put the highlighting data in the metadata, I think I'd do it by making a note of the location in `getVarType` and `getNameType` in `TTImp.Elab.App`.

EDIT: It's probably not necessary for this job, but it's also worth looking at `TTImp.Elab.Check` which has the top level type declarations and various support functions for elaborating terms. Things like creating metavariables for implicits, going into a new scope, calling the unifier and dealing with anything where we need backtracking.

abailly:
Thanks a lot! I will have a look ASAP but want to finish this PR first so that we can merge it without impacting anyone.

abailly:
@edwinb ready to merge!

christian-public:
@abailly I can confirm that this works in Atom (maybe you tested it already). Nice work on the IDE Mode/highlighting! :)

abailly:
Thanks a lot @chrrasmussen! It's only the beginning of the journey :)

edwinb:
Oh, I should have merged this by now, sorry. Thanks for the updates!

PR #206, opened 2020-03-01
Only ignore the executables in the root folder

christian-public:
The previous rule accidentally ignored all files in `tests/idris2`.

I found which rule that ignored the files in `tests/idris2` by using the command `git check-ignore -v <ignored-file>`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,8 @@
 
 *.dSYM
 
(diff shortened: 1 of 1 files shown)


edwinb:
Good idea, this keeps annoying me but never quite enough to fix it! Cheers.

PR #208, opened 2020-03-02
Simplify 'test' target in libs/network/Makefile

ska80:


edwinb:
I wonder if it's also possible to move this test to the normal test suite rather than doing it here?

PR #211, opened 2020-03-04
Add 'cwd' REPL command to show the current working directory

ska80:


src/Idris/REPL.idr
```diff
@@ -464,6 +464,7 @@ data REPLResult : Type where
   ErrorsBuildingFile : String -> List Error -> REPLResult
   NoFileLoaded : REPLResult
   ChangedDirectory : String -> REPLResult
+  WorkingDirectory : String -> REPLResult
```
abailly: Maybe we could use a single `CurrentDirectory` result constructor for both commands?

abailly (commented):
Minor simplification suggested

src/Idris/REPL.idr
ska80: Yes, I think it makes more sense.  I will do that.

src/Core/Context.idr
```diff
@@ -1746,6 +1746,13 @@ setWorkingDir dir
          cdir <- coreLift $ currentDir
          put Ctxt (record { options->dirs->working_dir = cdir } defs)
 
+export
+getWorkingDir : {auto c : Ref Ctxt Defs} -> Core String
+getWorkingDir
+    = do defs <- get Ctxt
+         cdir <- coreLift $ currentDir
+         pure cdir
```
clayrat: Can we just simplify the whole thing to `coreLift currentDir`? Doesn't seem like we even need to touch the context here.

ska80: > Can we just simplify the whole thing to `coreLift currentDir`? Doesn't seem like we even need to touch the context here.

Will do that :)

edwinb:
Looks a good idea to me, thanks.

PR #212, opened 2020-03-04
Remove rig count

andrevidela:
Hi, I have this branch on which I've been working in order to experiment with different semi rings for tracking linearity. It remove the access to the RigCount constructor and replace with interfaces and functions such as

```
-- check for linear
isRig1 : (Semiring a, Eq a) => a -> Bool

-- eliminator for semirings
elimSemi : (Semiring a, Eq a) => b -> b -> (a -> b) -> a -> b 

-- create the value for neutral addition which stands for "erased"
ring0 : (Semiring a) => a
```

Should this be merged onto master or should I keep running my own branch until I have more substantial results?

### DIFF
--- a/src/Algebra.idr
+++ b/src/Algebra.idr
@@ -1,0 +1,9 @@
+module Algebra
+
+import public Algebra.ZeroOneOmega
+import public Algebra.Semiring
+import public Algebra.Preorder
+
+public export
+RigCount : Type
+RigCount = ZeroOneOmega
--- a/src/Algebra/Preorder.idr
+++ b/src/Algebra/Preorder.idr
@@ -1,0 +1,29 @@
+module Algebra.Preorder
+
+||| Preorder defines a binary relation using the `<=` operator
+public export
+interface Preorder a where
+  (<=) : a -> a -> Bool
+  preorderRefl : {x : a} -> x <= x = True
+  preorderTrans : {x, y, z : a} -> x <= y = True -> y <= z = True -> x <= z = True
+
+||| Least Upper Bound, replace max using only Preorder
+export
+lub : Preorder a => a -> a -> a
+lub x y = if x <= y then y else x
+
+||| Greatest Lower Bound, replaces min using only Preorder
+export
+glb : Preorder a => a -> a -> a
+glb x y = if x <= y then x else y
+
+||| Strict less-than using the relation from a preorder
+public export
+(<) : (Preorder a, Eq a) => a -> a -> Bool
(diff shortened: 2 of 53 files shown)


edwinb:
It's probably a good idea to merge this sooner rather than later, although I should probably review it first to make sure I understand what's going on... I'll take a proper look in a bit.

src/Algebra/Nats.idr
```diff
@@ -0,0 +1,63 @@
...
+  constructor MkInfiniteRange
+  lo: Nat
+  up: Maybe Nat
+  lte : lo <|| up
+
+sumlte : (lte1 : (l1 <|| u1)) -> (lte2 : (l2 <|| u2)) -> ((l1 + l2) <|| (u1 + u2))
+sumlte LTEZero LTEZero = LTEZero
+sumlte LTEZero (LTESucc x) {u1=Nothing} = ?wat
+sumlte LTEZero (LTESucc x) {u1=Just y} = ?wat2
+sumlte (LTESucc x) lte2 = ?sumlte_rhs_2
```
keram: are these holes intentional? @andrevidela

andrevidela: No, some of those proofs need to go away anyway, sorry I should have put 'WIP' in the title since we've had some offline discussion with edwin to sort this out. I'll change the title and submit again once its ready

andrevidela:
Alright this is ready for review, here is a rundown of the changes:

- The constructors for `RigCount` have been hidden. Which means the only way to create new values of RigCount is by using the `Semiring` operations. I named the neutral for addition `erased` and the neutral for multiplication `linear` in order to stress the linear semantics. The addition and multiplication operators are `|+|` and `|*|` respectively
- The RigCount also has a `Preorder` that is bounded by `top` which acts as the default linearity.
- All pattern matching on RigCount is now impossible. However there is semiring eliminator called `elimSemi` which allows to eliminate a generic semiring.
- `RigCount` is now simply an alias for the type `ZeroOneOmega`

Notes:

- Some rules look a bit fishy and complicated, I've added a comment for most of them to highlight that there might be a better more general rule to implement them
- By refactoring the code I found multiple occurrences of `if isErased x then erased else top` which can be translated into semiring operations by `x |*| top`, does that make sense? I don't know

andrevidela:
@edwinb I don't understand why the print statements in chez don't seem to get executed. Do you have an idea why? I odn't see where this PR changes anything specific to IO

andrevidela:
@edwinb I rebased from master and now everything works as expected again

PR #213, opened 2020-03-05
Add getChar, putChar and putCharLn to prelude

ska80:


edwinb:
Good to have, thanks.

PR #215, opened 2020-03-06
Make BSD 'sed' happy

ska80:


edwinb:
Aha, thank you! I think the travis failure is due to something else, so this should be safe to merge.

Issue #187, opened 2020-01-24
Cannot compile prelude with latest Idris, file read error


andrevidela:
steps to reproduce (on Mac OS 10.15.2)

1. clone the project
2. `make install`

## Expected output

it works 

## Actual output

```
make -C libs/prelude install IDRIS2=../../idris2
../../idris2 --install prelude.ipkg
Uncaught error: File error (prelude): File Read Error
make[1]: *** [install] Error 1
make: *** [install-libs] Error 2
```

I can reproduce the problem with those steps as well:

1. compile Idris with `make idris2`
2. go to the prelude folder `cd libs/prelude`
3. compile manually with `../../idris2 --install prelude.ipkg`

The error is the same.

I checked my permissions are correct (even with `chmod 777`) and tried running as sudo and the result is the same.

Any ideas?

christian-public:
I believe this problem may occur if `make install` does not complete succesfully (for example if some tests fail).

`Uncaught error: File error (prelude): File Read Error` happens because it is looking for the `<idris2 install dir>/support/chez/support.ss`, which has not been installed yet. If you run `make install-support`, then it should be possible to install the libraries.

edwinb:
If this is to do with the Mac chez test failures, perhaps it's okay now?

andrevidela:
I tried again today and here are my findings:

It now fails at the `idris2-fromc` step, because `sed -i` isn't perfectly Multiplatform with Mac OS. I found this PR on Couch-db that fixes it by adding an empty string after the `-i` option when running on Mac OS.

Of course it works fine if I replace the `sed` call by `gsed` which is the gnu version but compiled for Mac OS.

However this fix only creates another, because compilation fail a bit later on compiling the c rts

```
make -C dist
make -C rts
clang  -O2 -Wall -std=c99 -pipe -fdata-sections -ffunction-sections -D_POSIX_C_SOURCE=200809L  -I/usr/local/include  -DIDRIS_TARGET_OS="\"darwin\"" -DIDRIS_TARGET_TRIPLE="\"x86_64-apple-darwin19.3.0\"" -fPIC   -c -o idris_rts.o idris_rts.c
idris_rts.c:1163:21: error: variable has incomplete type 'struct timespec'
    struct timespec t;
                    ^
idris_rts.c:1163:12: note: forward declaration of 'struct timespec'
    struct timespec t;
           ^
idris_rts.c:1167:12: warning: implicit declaration of function 'nanosleep' is
[...]

PR #216, opened 2020-03-06
Fix sed to work with macOS

andrevidela:
A small fix for the first bit of #187


edwinb:
Ah,I see two fixes for this came in at once! Thanks anyway :).

PR #220, opened 2020-03-08
Add Text.{Lexer,Parser}.Core to the list of installed modules

ziman:
Otherwise these files do not get installed, which leads to missing-module errors on new installations and stale library modules on existing installations.


ohad:
Ooops! Sorry about that --- thanks!

Issue #218, opened 2020-03-07
Can't build latest on OSX


shmish111:
macOS: 10.14.6
idris: 1.3.2
Idris2: https://github.com/edwinb/Idris2/commit/34044597c07a96235f8b54727f7655e770727c57

```bash
➜  Idris2 git:(3404459) ✗ make idris2
Using Idris 1 version: 1.3.2
Building Idris 2 version: 0.0.0-34044597c
idris --build idris2.ipkg
Entering directory `./src'
Type checking ./Idris/Main.idr
./Idris/Main.idr:51:15-21:
   |
51 |               Nothing => setPrefix yprefix
   |               ~~~~~~~
When checking right hand side of Main.case block in updatePaths at ./Idris/Main.idr:49:15-21 with expected type
        Core ()

Can't disambiguate name: Main.yprefix, YafflePaths.yprefix

./Idris/Main.idr:205:7-206:19:
    |
205 |     = do putStrLn yprefix
    |       ~~~~~~~~~~~~~~~~~~~ ...
When checking right hand side of quitOpts with expected type
        IO Bool

Can't disambiguate name: Main.yprefix, YafflePaths.yprefix

make: *** [dist/idris2.c] Error 1
```

I tried installing the latest idris 1 but that failed to build too 😄 
```bash
➜  Idris-dev git:(bbd0f286c) ✗ stack build
idris> build (lib + exe)
Preprocessing library for idris-1.3.2..
Building library for idris-1.3.2..
Preprocessing executable 'idris-codegen-c' for idris-1.3.2..
Building executable 'idris-codegen-c' for idris-1.3.2..
Preprocessing executable 'idris-codegen-javascript' for idris-1.3.2..
Building executable 'idris-codegen-javascript' for idris-1.3.2..
Preprocessing executable 'idris' for idris-1.3.2..
Building executable 'idris' for idris-1.3.2..
Preprocessing executable 'idris-codegen-node' for idris-1.3.2..
Building executable 'idris-codegen-node' for idris-1.3.2..
Building libraries...
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C prelude build
../../.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/idris/idris --build prelude.ipkg
Type checking ./Prelude/Nat.idr
idris: No trivial solution
make[1]: *** [build] Error 1
make: *** [build] Error 2

--  While building package idris-1.3.2 using:
[...]

PR #224, opened 2020-03-09
Fix build on macOS

christian-public:
As suggested by @edwinb in #218, adding `#include <time.h>` does fix the build on macOS!

Hopefully the CI will confirm that it still works on Linux.


edwinb:
Thanks!

Issue #229, opened 2020-03-12
Matching on erased arguments within an erased function


ohad:
# Steps to Reproduce
```
$ cat Case.idr 
import Decidable.Equality

0 foo : (i, j : Nat) -> Bool
foo i j = case decEq i j of
            Yes pf => True
            No  pf => False
```

# Expected Behavior
```
$ idris2 -c --no-banner Case.idr 
1/1: Building Case (Case.idr)
```

# Observed Behavior
```
$ idris2 -c --no-banner Case.idr 
1/1: Building Case (Case.idr)
Case.idr:4:11--7:1:While processing right hand side of Main.foo at Case.idr:4:1--7:1:
Attempt to match on erased argument Yes pat0: in Main.case block in 1507(292)
```

Since `foo` is itself irrelevant, I'd expect the irrelevance to extend into any case-splits it performs.

ohad:
BTW, until this issue is fixed, I can work around it using a `with` block:
```
$ cat CaseWith.idr 
import Decidable.Equality

0 foo : (i, j : Nat) -> Bool
foo  i j with (decEq i j)
 foo i j | Yes pf = True
 foo i j | No  pf = False
$ idris2 -c --no-banner CaseWith.idr 
1/1: Building CaseWith (CaseWith.idr)
```

ziman:
For some reason, `case` elaboration produces a lifted function with a zero-quantified scrutinee. With `:log 5`, I get:
```idris
LOG 2: Case function type: Main.case block in 1500(292): ((j : Prelude.Nat) -> ((i : Prelude.Nat) -> ((0 {scr:291} : (Prelude.Dec (Builtin.=== Prelude.Nat i[0] j[1]))) -> Prelude.Bool)))
```
Could that be a heuristic that works well for linear values but probably not so well for erased values?

Either way, this is an issue in elaboration, not in the typing rules.

(A small technical note: `foo` is not irrelevant, it's just marked as erasable. "Irrelevance" is usually used for things that are not only erasable at runtime but also ignored in equality.)

ziman:
The case elaborator uses `caseRig` to bind the scrutinee.
```idris
let casefnty =
       abstractFullEnvType fc (allow splitOn env)
           (maybe
               (Bind fc scrn
                   (Pi caseRig Explicit scrty)
                   (weaken caseretty))
               (const caseretty)
               splitOn)
```
In this case, `caseRig = Rig0`, so we observe the behaviour described in the OP.

[...]

PR #230, opened 2020-03-13
Compute case scrutinee quantity more precisely.

ziman:
Should solve #229. The whole stdlib builds, all tests pass, and [Ohad's example program](https://github.com/edwinb/Idris2/issues/229) passes the compiler now, too.

I'm not sure that this patch computes the result in the most efficient way that the internal API provides, though.


ziman:
Should we add Ohad's program as a test? Plus perhaps a few other corner cases, such as `Refl` and such.

ziman:
It seems I misunderstood what `caseRig` means. Let's not merge this just yet.

PR #231, opened 2020-03-13
Add version and file to TTCError

andrevidela:
I had some issues while playing with TTC files and couldn't tell which version from which file wasn't matching. So I've added this debug info to the `Format` constructor


edwinb:
Good idea, thanks

PR #235, opened 2020-03-18
bugfix: update test expected output

ohad:
test block number changed between versions


ohad:
OK, I ran it locally with a stale copy of the library binaries. Going to cancel the PR.

PR #237, opened 2020-03-22
Literate programming

petithug:
- Add support for .lidr files.
- Add support for Case Split.
- Add new Parse Error for Literate fails.
- Add support for Add Clause.
- Add support for Add Lemma.
- Literate processing fails if code blocks are not separated by a blank
  line.


petithug:
I am also preparing a patch for idris2-vim, and I found some issues, so you may want to wait until I fix them (which will probably be next week-end).  Meanwhile comments and suggestions are welcome.

edwinb:
Thank you! I know some other people have been wanting this too. I'm happy to merge now, unless you'd like me to wait for the idris2-vim updates.

petithug:
I found 4 bugs as I was modifying idris2-vim, so I think it's simpler to wait until I update the PR this week-end.  I'll send the PR for idris2-vim at the same time.

jfdm:
This is a good addition to Idris2, I am partial to literate modes and honestly you ticked off one of my own TODO items when looking to contribute to Idris2.

I wonder if at some point we should look at using [`Text.Literate`](https://github.com/idris-lang/Idris-dev/blob/master/libs/contrib/Text/Literate.idr) from Idris' `contrib` library instead of keeping the literate mode of 'Bird Style'. Of note this suggestion should not block this PR!

Such a change should make the literate mode more versatile and allow multiple styles to be supported. Definitely once PR #4834 on Idris has bubbled into a tagged version, as I recently fixed a stupid mistake by the library's author.

petithug:
In fact only one of the bugs I found was in that patch, and I fixed it so it is good to be merged.
The other bugs were either non-implemented features or bugs in other parts of Idris2.  I opened bug reports for the latter.

edwinb:
Looks good, thank you!

PR #239, opened 2020-03-27
Add basic help to REPL

maxdeviant:
:wave:

This PR adds basic help to the REPL.

I based much of the implementation off of the help system for the Idris 1 REPL, with some tweaks here and there to account for things that work in Haskell but not in Idris.

The majority of the commands have been ported, but I'm still working on some of the more advanced ones. Where possible, I took the help descriptions from the Idris 1 REPL, but 1) I'm not sure if those are all entirely accurate, and 2) there are some commands in Idris 2 not present in Idris 1. I'll probably need some help documenting those.

This is my first time working with Idris, so any feedback on areas that could be improved would be very welcome :slightly_smiling_face:

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -1519,56 +1519,76 @@ Show CmdArg where
   show NameArg = "<name>"
   show ExprArg = "<expr>"
 
+export
+data ParseCmd : Type where
+     ParseREPLCmd : List String -> ParseCmd
+     ParseKeywordCmd : String -> ParseCmd
+     ParseIdentCmd : String -> ParseCmd
+
 CommandDefinition : Type
 CommandDefinition = (List String, CmdArg, String, Rule REPLCmd)
 
 CommandTable : Type
 CommandTable = List CommandDefinition
 
-eatCommand : List String -> Rule ()
-eatCommand names = do
-  symbol ":"
-  replCmd names
+extractNames : ParseCmd -> List String
+extractNames (ParseREPLCmd names) = names
+extractNames (ParseKeywordCmd keyword) = [keyword]
+extractNames (ParseIdentCmd ident) = [ident]
 
-noArgCmd : List String -> REPLCmd -> String -> CommandDefinition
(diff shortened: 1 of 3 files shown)


maxdeviant:
After a bit more work, I think I'm happy with the state of the PR and am ready for it to be reviewed.

All of the named commands have been added to the `CommandTable`, which is used to recognized the REPL commands and add them to the help. I also added proper spacing to the help table so that it is nicer to look at.

Here is the help in action:

image

ska80:
> After a bit more work, I think I'm happy with the state of the PR and am ready for it to be reviewed.

Good job!  I think you have missed `:cd ` command that changes the current working directory.

edwinb:
Looks great, thank you! I'll merge shortly. One small thing is that in Idris 1, ':?' also works for help, but I can add that after merging.

maxdeviant:
> Good job! I think you have missed `:cd ` command that changes the current working directory.

It doesn't look like the `:cd` command is in Idris 2 yet. I just ported the existing commands and added `:help`. I'd be happy to look into adding more commands though.

maxdeviant:
> Looks great, thank you! I'll merge shortly. One small thing is that in Idris 1, ':?' also works for help, but I can add that after merging.

Thank you! I did try to add `:?` for the help, but it seemed to not fit well in the current command-parsing architecture. It looks like the parser treats `:?` as a single symbol token:

> Main> :?
> Parse error: Expected 'case', 'if', 'do', application or operator expression (next tokens: [symbol :?, end of input])

However, we currently always eat a `:` symbol first when parsing a REPL command. I can look to add this, unless you get to it first :slightly_smiling_face:

ska80:
> > Good job! I think you have missed `:cd ` command that changes the current working directory.
> 
> It doesn't look like the `:cd` command is in Idris 2 yet. I just ported the existing commands and added `:help`. I'd be happy to look into adding more commands though.

You may take a look at ``REPLCmd`` type, it does have ``:cd`` REPL command :)

There are also ``:load``/``:l`` and other commands that could be documented.

PR #243, opened 2020-03-28
Add `intercalate` for `Data.List`

maxdeviant:
I noticed the absence of `intercalate` while working on #239.

While I didn't need it at the moment, it seems like something that would be useful to have in `Prelude`.

Implementation was ported from Idris 1.


edwinb:
Thanks! Yes, this is a good addition.

PR #250, opened 2020-03-29
Fix runtime error in `the Int (cast "")`

ziman:
With the following program, Idris errors out.
```idris
$ cat x.idr
main : IO ()
main = printLn (the Int (cast ""))
```
```
$ idris2 x.idr -x main
Exception in string-ref: 0 is not a valid index for ""
```

It turns out that the RTS always attempts to look at first character of the string, even if the string is empty.

This patch fixes it:
```
$ idris2 x.idr -x main
0
```

### DIFF
--- a/support/chicken/support.scm
+++ b/support/chicken/support.scm
@@ -19,8 +19,10 @@
     (if (number? x) x 0)))
 (define destroy-prefix
   (lambda (x)
-    (if (eqv? x "") ""
-      (if (eqv? (string-ref x 0) #\#) "" x))))
+    (cond
+      ((equal? x "") "")
+      ((equal? (string-ref x 0) #\#) "")
+      (else x))))
 (define cast-string-int
   (lambda (x)
     (floor (cast-num (string->number (destroy-prefix x))))))
--- a/support/chicken/support.scm
+++ b/support/chicken/support.scm
@@ -19,7 +19,8 @@
     (if (number? x) x 0)))
 (define destroy-prefix
   (lambda (x)
-    (if (eqv? (string-ref x 0) #\#) "" x)))
+    (if (eqv? x "") ""
+      (if (eqv? (string-ref x 0) #\#) "" x))))
 (define cast-string-int
   (lambda (x)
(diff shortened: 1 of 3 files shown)


support/chicken/support.scm
```diff
@@ -19,7 +19,8 @@
     (if (number? x) x 0)))
 (define destroy-prefix
   (lambda (x)
-    (if (eqv? (string-ref x 0) #\#) "" x)))
+    (if (eqv? x "") ""
+      (if (eqv? (string-ref x 0) #\#) "" x))))
```
keram: Hello @ziman :),
What you think instead of nested `if` use something like: 
```lsp
(define destroy-prefix
  (lambda (x)
    (cond
     (eqv? x "") ""
     (eqv? (string-ref x 0) #\#) ""
     (else x))))
``` 
or
```lsp
(define destroy-prefix
  (lambda (x)
    (if (or (eqv? x "") (eqv? (string-ref x 0) #\#))
      ""
      x)))
```
Should be equivalent and i feel it reads slightly better although haven't tested myself.

david-christiansen:
`eqv?` is not the right predicate to compare strings, because it's implementation-defined in Scheme whether two strings with the same contents but that were allocated separately are equated. For this, `equal?` or `string=?` would be the right predicate, or testing the length of the string.

david-christiansen:
Stylistically, the `cond` looks more idiomatic to me, as well.

support/chicken/support.scm
ziman: o/ :)

Thank you!

ziman:
I wanted to be as conservative as possible here because I don't have much clue about what's portable in Scheme and what's not. And still I did not get it right, but fortunately there are clever people out there, thanks a lot! :)

edwinb:
I've kept getting the equalities in Scheme wrong too... I wouldn't be surprised if some errors still linger in the support code. Anyway, thanks for fixing, and for all your other patches lately!

There may be a couple of other places in the runtime where Scheme throws an exception we're not dealing with properly, especially in IO code.

PR #251, opened 2020-03-31
Show a hint for help on the banner

ska80:


edwinb:
Ah yes, good idea, thanks!

PR #253, opened 2020-04-01
[ fix ] properly nesting multiline comments

gallais:
We should be able to comment out any part of a program that parses
and still get a program that parses. This means we need to properly
ignore `{-` and `-}` when hidden behind a `--` comment or inside of
a string literal.


edwinb:
Excellent, thanks. I wrote this bit on a train and I think I never came back to polish it... so I hadn't even spotted that was an issue!

PR #255, opened 2020-04-01
Don't clobber CWD in `--exec`

ziman:
If you want to open a file in an Idris2 program, you need to use its absolute path because `--exec` chdirs somewhere else to exec the Scheme. This patch fixes that.

This is mostly an attempt to get it done and move on so if you prefer to have something cleaner, I can have another go at it.

### DIFF
--- a/src/Compiler/Common.idr
+++ b/src/Compiler/Common.idr
@@ -21,10 +21,10 @@ public export
 record Codegen where
   constructor MkCG
   ||| Compile an Idris 2 expression, saving it to a file.
-  compileExpr : Ref Ctxt Defs ->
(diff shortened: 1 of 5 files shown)


ziman:
Maybe the best thing would be to build everything somewhere in `/tmp` and just copy the resulting binary (or Scheme script) to the path given by `-o`?

edwinb:
In general it'd be nice to have just one file generated, but at least for Chez there is no easy way (to my knowledge) of generating a single executable, which is why I added the startup script. (There are ways, but I don't know how stable they are...) Anyway, this is a useful fix, thanks.

PR #259, opened 2020-04-02
A little cleanup

gallais:


gallais:
I have tried to cleanup the definition of `fspanEnd` by encapsulating the
calls to `prim__*` behind an auxiliary `strUncons` but we would still need
an `assert_total` because the relationship between `str` and the tail returned
by `strUncons` is not evident. So I gave up instead (well... I looked into the
`strM` view too and that also led to a mess so then I gave up).

PR #261, opened 2020-04-03
Resugar `Nat` literals

gallais:


edwinb:
Thanks for this! mapPTermM will probably be useful a few more times...

PR #268, opened 2020-04-06
[ new ] interactive test runner

gallais:
New interactive mode for the test runner.

* Print diff in case of mismatch (using `git diff`)
* Ask user whether they want to overwrite the current golden value

New options syntax for the test runner.

* `--interactive` flag
* `--only` should be the *last* flag and followed by a *list* of cases of interest. As before, if the list is empty we run all the tests.

### DIFF
--- a/.ci-dockerfiles/ci-run.dockerfile
+++ b/.ci-dockerfiles/ci-run.dockerfile
@@ -4,4 +4,4 @@ copy . /Idris2-dev
 workdir /Idris2-dev
 run make idris2
 run make libs
-run make test
+run make test INTERACTIVE=''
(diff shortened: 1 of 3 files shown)


jfdm:
Nice! 

Thanks for this, it will help things out greatly.

I was thinking, could we get a flag `--accept` to provide a default yes answer for accepting new tests.

gallais:
> I was thinking, could we get a flag `--accept` to provide a default yes answer for accepting new tests.

Running `yes y | make test` should do the trick.

jfdm:
> Running `yes y | make test` should do the trick.

Yes, indeed it will. I was thinking that `--accept` would be that little bit better and not reliant on having `yes`.

PR #270, opened 2020-04-06
[ refactor ] Avoid lookup + insert in IntMap's mergeWith

gallais:
If we're happy with this, I can port the changes to all the other map
variants in `Data/` (Related: it would be nice to have some staging
capabilities).

### DIFF
--- a/src/Data/IntMap.idr
+++ b/src/Data/IntMap.idr
@@ -192,15 +192,21 @@ treeDelete {n=(S (S _))} k (Branch3 t1 k1 t2 k2 t3) =
       Left t3' => Left (Branch3 t1 k1 t2 k2 t3')
       Right t3' => Left (merge3 t1 k1 t2 k2 t3')
 
+treeFoldl : (acc -> (Key, v) -> acc) -> acc -> Tree n v -> acc
+treeFoldl c n (Leaf k v) = c n (k, v)
+treeFoldl c n (Branch2 t1 _ t2) = treeFoldl c (treeFoldl c n t1) t2
+treeFoldl c n (Branch3 t1 _ t2 _ t3) =
+  treeFoldl c (treeFoldl c (treeFoldl c n t1) t2) t3
+
+treeFoldr : ((Key, v) -> acc -> acc) -> acc -> Tree n v -> acc
+treeFoldr c n (Leaf k v) = c (k, v) n
+treeFoldr c n (Branch2 t1 _ t2) =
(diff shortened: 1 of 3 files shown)


edwinb:
I hope we'll be able to have better staging capabilities in Idris 2 - it's annoying to have to do this but it does have an impact on performance (I can't remember how much though, should have written it down at the time!)

Idris 1 is happy to specialise functions by implementation of an interface, but it doesn't specialise data constructors which is what we'd need here. We should look at that later.

edwinb:
I'm trying to find out what effect this has on performance, but as far as I can tell any difference between this way and the old way is just noise caused by whatever my browser has decided to do in the background, or something... I'll investigate properly tomorrow.

gallais:
I have ported the changes to StringMap. I can't really see a difference on
`make test` but then again I expect we are not generating huge maps of
meta-variables & the like.

I will try it on self-hosted and/or frex.

gallais:
Alright, here is the before and after on Idris2-SelfHost after running
`time ../../Idris2/idris2 --check Idris/REPL.idr` 3 times for each build:

* Before ~3m53s (with one outlier)

```
real	4m38.194s
user	3m53.068s
sys	0m44.804s

real	4m37.368s
user	3m53.620s
sys	0m43.584s

real	4m47.731s
user	4m1.772s
sys	0m45.612s
```

* After ~3m49s

```
real	4m33.859s
user	3m49.672s
sys	0m44.072s

real	4m33.544s
user	3m49.096s
sys	0m44.300s

real	4m34.575s
user	3m49.896s
sys	0m44.560s
```

gallais:
Porting the changes to `NameMap` leads to worse compile times.
I am going to try to revert either the `StringMap` or the `IntMap`
and see whether each is contributing positively or not.


```
real	4m43.973s
user	3m57.836s
sys	0m45.912s

real	5m2.794s
user	4m13.060s
sys	0m48.300s

real	5m8.913s
user	4m17.520s
sys	0m49.956s
```

Reverting that last attempt and going back to `IntMap` + `StringMap` gives me
compile times of ~4m instead of the ~3m49s I had earlier. I don't understand
what is happening. Clearly the numbers are not super reliable...

edwinb:
Finally catching up... I wonder if the closure in the higher order function is more costly than you'd expect here? These operations are used a lot, especially the NameMap, which helps turns human readable names into resolved names, so any small difference might add up.

Issue #269, opened 2020-04-06
Record, parameters, interface, and implementation declarations ignore `%unbound_implicits off`


ohad:
# Steps to Reproduce
```
$ cat Records.idr
import Data.Vect

%unbound_implicits off

record Foo (x : Vect n Nat) where
  constructor MkFoo

parameters (Foo : Vect n Nat)
  bar : Nat
  bar = 0


interface Foo (a : Vect n Nat) where
  baz : Nat
  
implementation Functor (Vect n) where
$ idris2 --no-banner -c Records.idr
```
# Expected Behavior
Some kind of error, as `n` is not bound anywhere. Cf.:
```
$ cat Records.idr 
import Data.Vect

%unbound_implicits off

Foo : (x : Vect n Nat) -> Type
Foo x = Nat

$ idris2 --no-banner -c Records.idr 
1/1: Building Records (Records.idr)
Records.idr:5:17--5:19:While processing type of Foo at Records.idr:5:1--6:1:
Undefined name n
Records.idr:6:1--8:1:No type declaration for Main.Foo
```
# Observed Behavior
```
$ idris2 --no-banner -c Records.idr 
1/1: Building Records (Records.idr)
$
```

ohad:
The issue is that the `%unbound_implicits off` pragma is processed at the `TT` level, whereas the decision whether to bind the implicit names is taken during desugaring.

ohad:
Fixed in #271

PR #271, opened 2020-04-06
Bugfix #269

ohad:
Fix for #269 

Make desugaring/elaboration of interfaces, interface implementations,
records, and parameter blocks take into account the pragma
`%unbound_implicits off`.

Main changes:

(a) Execute the pragma also during desugaring
(b) Check whether `isUnboundImplicits` is on at each desugaring step

Alternatives I didn't take:

(1) Changing `findBindableNames` to effectfully check the flag.

  Rationale:
  Apart from turning a pure function into an effectful one, this
  would mean repeatedly calling `findBindableNames`, only to do
  nothing once the flag is read.

(2) Adding another function that takes multiple places (list of terms)
  that might contain bindable names, and before dispatching
  `findBindableNames` on each term, checking the flag.

  Rationale: I didn't want to add another abstraction, as that would
  also require further changes inside TTImp. (weak rationale)

@edwinb @gallais : if you prefer (2), I can do that.

### DIFF
--- a/src/Idris/Desugar.idr
+++ b/src/Idris/Desugar.idr
@@ -723,7 +723,7 @@ mutual
                         then concatMap (findBindableNames True
                                          (ps ++ fnames ++ map fst params) [])
                                        (map snd params')
-                     else []
+                        else []
            fields' <- traverse (desugarField (ps ++ fnames ++ map fst params))
                                fields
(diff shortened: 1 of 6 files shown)


gallais (commented):
I think you are allowed to keep the `let`-bindings even if you use `!`
to perform an effectful computation in the let-bound expression.

ohad:
Thanks @gallais . I'll update that.

PR #272, opened 2020-04-07
Add equality functions for Vectors of potentially different lengths.

jfdm:
Sized lists are useful, however, interfaces require that the length value is fixed.
The added functions allow for vectors of potentually different lengths to be compared.


jfdm:
Obviously something has changed since I wrote this, as the tests are failing. I will reopen once I have addressed the impact.

PR #277, opened 2020-04-13
Latest rtd theme for docs

jeetu7:
Seems working with latest rtd theme and sphinx 3.0.1(installed automatically by rtd theme).

Changes:

1. Seperate build and source directory.
2. proof-index label warning resolved.
3. Most of the warnings at the time of make html resolved.
4. Refs and footnotes are named so not to be issue.

Issues:

1. footnotes and refs are not showing proper (going to the next line). (Will solve it eventually)
2. Navigation buttons should be displayed on the top and bottom as per rtd theme config. But not working. Still only available at the bottom.


edwinb:
Thanks for the updates!

PR #282, opened 2020-04-14
provide building blocks for binary network I/O

abailly:
This provides some low-level machinery to make it possible to do binary I/O using network library:

* expose low-level free/malloc functions for byte buffers
* provide basic peek/poke functions to operate on buffers

### DIFF
--- a/libs/network/Network/Socket/Raw.idr
+++ b/libs/network/Network/Socket/Raw.idr
@@ -23,7 +23,18 @@ data SockaddrPtr = SAPtr AnyPtr
 
 -- ---------------------------------------------------------- [ Socket Utilies ]
 
+||| Put a value in a buffer
+export
+sock_poke : BufPtr -> Int -> Int -> IO ()
+sock_poke (BPtr ptr) offset val = cCall () "idrnet_poke" [ptr, offset, val]
+
+||| Take a value from a buffer
+export
+sock_peek : BufPtr -> Int -> IO Int
+sock_peek (BPtr ptr) offset = cCall Int "idrnet_peek" [ptr, offset]
+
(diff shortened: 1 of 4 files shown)


edwinb:
Is this related to `Data.Buffer` somehow? I wonder if it'd be better to use/refine that rather than implement something similar separately.

edwinb:
Although... currently `Data.Buffer` is using the scheme primitives, and the network code isn't, so that suggests some kind of deeper thought is required on how to tie all of this together.

abailly:
I agree it would be better to have a single solution for managing bytes buffers, the problem is that the network does not use scheme primitives, hence it is hard to pass `Data.Buffer` back and forth because they don't have the same format. I tried that route and failed. 
A possible option would be to enhance `Data.Buffer` to allow building `Buffer` instances directly from C buffers.

abailly:
I have been working a bit on reading/writing binary encoded data from/to network and there's a lot of overlap with `Data.Buffer`, of course, plus a lot of low-level plumbing that makes it extremely tedious. I was thinking of writing something like the [binary](http://hackage.haskell.org/package/binary) Haskell package but without a unified view on what an array of bytes is, that will be useless.
Ideally, we would settle on `Data.Buffer` and make the network package use it for "raw" bytes array writing and reading but I think this requires changes to the `Data.Buffer` package to cope with passing raw byte arrays to/from C space. WDYT @edwinb  ?

edwinb:
I think we can delay doing anything about buffers for now, and merge this is as. I have a half baked plan to move as much of the support code as possible to C libraries, which would help here, but I don't know when I'll get around to it.

PR #284, opened 2020-04-15
use quotient on scheme backends

abailly:
We compile divisions of Integers to quotient instead of '/' and add some tests for numbers, as using `/` has the undesirable side-effect of producing a non integral number


ziman:
Related <s>(and probably conflicting)</s>: https://github.com/edwinb/Idris2/pull/281

ziman:
If you have a clue what's going on in https://github.com/edwinb/Idris2/pull/281#issuecomment-613673042 and how to fix it, that would be great.

abailly:
@ziman  Ah, did not check that PR, sorry if this PR is conflicting. I will have a look

ziman:
It's actually not conflicting, as far as I can tell! It actually fixes a related issue that my PR missed.

edwinb:
Thanks!

PR #287, opened 2020-04-15
Fix Idris1 download link

Porges:
This was pointing to the wrong place.


edwinb:
Ah yes, I missed that when I updated the site. Thanks for the fix.

PR #290, opened 2020-04-15
Refactor Parser.Lexer.{ident, OpCode} for reuse in IDEMode

fabianhjr:
Related Diff: https://github.com/edwinb/Idris2/commit/1a4f42425963cc414997d33fc9f4b8bac9fc1fbf#diff-5e7cbe76487381054be3ce2d96369916

Reusing the same function from the Idris Parser will avoid any future out of sync issues when the ident lexer changes. Also some cleanup of opChar.

There was an out of sync issue where identifiers that idris accepted like `álo` wouldn't be recognized by IDEMode.

### DIFF
--- a/src/Core/Name.idr
+++ b/src/Core/Name.idr
@@ -56,7 +56,7 @@ showSep sep (x :: xs) = x ++ sep ++ showSep sep xs
 ||| Check whether a given character is a valid identifier character
 export
 identChar : Char -> Bool
-identChar x = isAlphaNum x || x == '_' || x == '\''
+identChar x = isAlphaNum x || x == '_' || x == '\'' ||  x > chr 127
 
 export Show Name where
   show (NS ns n) = showSep "." (reverse ns) ++ "." ++ show n
--- a/src/Idris/Desugar.idr
+++ b/src/Idris/Desugar.idr
@@ -72,11 +72,10 @@ toTokList (POp fc opn l r)
          let op = nameRoot opn
          case lookup op (infixes syn) of
               Nothing =>
-                let ops = unpack opChars in
-                    if any (\x => x `elem` ops) (unpack op)
-                       then throw (GenericMsg fc $ "Unknown operator '" ++ op ++ "'")
-                       else do rtoks <- toTokList r
-                               pure (Expr l :: Op fc opn backtickPrec :: rtoks)
+                  if any isOpChar (unpack op)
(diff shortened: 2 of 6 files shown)


src/Core/Name.idr
```diff
@@ -56,7 +56,7 @@ showSep sep (x :: xs) = x ++ sep ++ showSep sep xs
 ||| Check whether a given character is a valid identifier character
 export
 identChar : Char -> Bool
-identChar x = isAlphaNum x || x == '_' || x == '\''
+identChar x = isAlphaNum x || x == '_' || x == '\'' ||  x > chr 127
```
gallais, line 60: Is this change unrelated and accidentally included from the unicode branch?

fabianhjr: No, the ident parser/lexer accepts chr > 127 for idents but Core.Name rejected them.

fabianhjr: I could write a test to make sure this is necesarry. BRB testing.

fabianhjr: Ok, couldn't figure out a quick way to make a test case however in TTImp.Interactive.GenerateDef:

```
fnName : Bool -> Name -> String
fnName lhs (UN n)
    = if any (not . identChar) (unpack n)
         then if lhs then "(" ++ n ++ ")"
                     else "op"
         else n
fnName lhs (NS _ n) = fnName lhs n
fnName lhs (DN s _) = s
fnName lhs n = show n
```

Here is where it would be used and as far as I understand this would misidentify unicode identifiers (such as ällo) as operators and that would mess with this:

```
         rhshole <- uniqueName defs [] (fnName False n ++ "_rhs")
```

fabianhjr: I think it would be better to remove `identChar` and use `identStrict` in its place.

gallais: :+1:  to using the same definition in both places if we want them to agree.

edwinb:
Thanks for the update. Should I wait for the `identChar/identStrict` update, or just merge now? I think this is okay to go as is, but if it won't be long I'll wait and check in again shortly.

fabianhjr:
I don't currently have time to quickly do the other change. It could be done in another pr.

edwinb:
Righto, thanks, in that case I'll just merge (although github seems sluggish at the moment).

PR #292, opened 2020-04-16
Fully abstract records

ohad:
Add support for implicit and auto implicit parameters to records, with rig annotations.

This allows us to declare records such as:
```idris
data NotZero : Nat -> Type where
  Is : {n : Nat} -> NotZero (S n)

record Positive (n : Nat) {auto 0 pos : NotZero n} where
  constructor Check

a : Positive 1
a = Check

b : Positive 2
b = Check
```
(taken from the added `test/idris2/record003`)

I'm not sure anyone would want the irrelevance annotation in the type, but we might as well allow it.

This is my first independent tweak to (small parts of) the elaborator, so I'd be grateful for a thorough review.

The refactoring of the parser's `pibindList` through `pibindListName` might slow it down a little bit. I haven't profiled to see if it's substantial. If it is, we can inline `pibindListName` and add a `Just`.


src/Idris/Parser.idr
```diff
@@ -1263,16 +1268,39 @@ fieldDecl fname indents
...
+         symbol "}"
+         pure $ map (\(c, n, tm) => (n, c, Implicit, tm)) params
+  <|> do symbol "{"
+         keyword "auto"
+         commit
+         start <- location
+         params <- pibindListName fname start indents
+         symbol "}"
+         pure $ map (\(c, n, tm) => (n, c, AutoImplicit, tm)) params
+  <|> do start <- location
```
gallais, line 1292: You could write

```idris
 <|> do symbol "{"
         info <- (AutoImplicit <$ keyword "auto" <|> pure Implicit)
         commit
         start <- location
         params <- pibindListName fname start indents
         symbol "}"
         pure $ map (\(c, n, tm) => (n, c, info, tm)) params
```

src/Idris/Syntax.idr
```diff
@@ -884,6 +884,12 @@ mapPTermM f = goPTerm where
       (::) . (\ c => (a, b, c)) <$> goPTerm t
                                 <*> go3TupledPTerms ts
 
+    go4TupledPTerms : List (a, b, c, PTerm) -> Core (List (a, b, c, PTerm))
+    go4TupledPTerms [] = pure []
+    go4TupledPTerms ((a, b, c, t) :: ts) =
+      (::) . (\ d => (a, b, c, d)) <$> goPTerm t
+                                   <*> go4TupledPTerms ts
```
gallais: Isn't there a `PTerm` potentially hidden in the `PiInfo` too?

gallais (changes requested):
Not sure why so many `commit`s had to be removed in the parser. Is it because
of `{ ...` vs. `{ auto ...`? In this case the suggested refactoring should take
care of that.

src/Idris/Syntax.idr
ohad: Yes, well spotted, thanks. 
Will fix.

We should double-check with @edwinb about the status of default implicit arguments. I think at some point he wanted to move away from them in Idris2 and use `auto` implicits.

ohad: Fixed. I'll poke @edwinb on slack

src/Idris/Parser.idr
ohad: Done, thanks.
Idris needed a little bit of help with type annotations.

ohad:
Ah, since @edwinb says we will support defaults, I'll add support for them too (should amount to just parsing + testing, but who knows up front)

ohad:
OK, I've added default arguments to record types, so this is ready to be merged.

(There's still a missing feature in allowing _fields_ to be auto / default implicits, but I'll leave that to the future.)

gallais (approved):

ziman:
I'm wondering about the auto implicit parameters (not fields).

Just to illustrate what I mean: in the case of `Positive`, I think I'd go for the following instead:
```idris
record Positive (n : Nat) where
  constructor Check
  {auto 0 isPositive : NotZero n}
```
Otherwise you wouldn't be able to even _talk about_ the positivity of numbers that have not already been proven to be positive.

I understand that `Positive` might be just a contrived minimal example to avoid a lot of explanation. Besides, even my suggestion does not parse currently, so it's certainly good to add that.

But I still wonder if it's ever useful to have an auto implicit as a parameter of a _type_ constructor.

src/Idris/Desugar.idr
```diff
@@ -691,20 +691,26 @@ mutual
           = map (\x => (Nothing, x)) (pairToCons p)
 
   desugarDecl ps (PImplementation fc vis pass is cons tn params impname nusing body)
-      = do is' <- traverse (\ ntm => do tm' <- desugar AnyExpr ps (snd (snd ntm))
-                                        pure (fst ntm, fst (snd ntm), tm')) is
-           cons' <- traverse (\ ntm => do tm' <- desugar AnyExpr ps (snd ntm)
-                                          pure (fst ntm, tm')) cons
+      = do is' <- traverse (\ (n,c,tm) => do tm' <- desugar AnyExpr ps tm
+                                             pure (n, c, tm')) is
+           let _ = the (List (Name, RigCount, RawImp)) is'
```
ziman, line 696: Interesting, are these `let _ = the T x` forms type declarations? Are you adding them to aid humans or the type checker? Should `<-` have a way of declaring these types?

ohad:
> I understand that `Positive` might be just a contrived minimal example to avoid a lot of explanation. 

Yes, it's a contrived example :). 


> I'm wondering about the auto implicit parameters (not fields).
> 
> Just to illustrate what I mean: in the case of `Positive`, I think I'd go for the following instead:
> 
> ```idris
> record Positive (n : Nat) where
>   constructor Check
>   {auto 0 isPositive : NotZero n}
> ```

Yes, that's a much better way to do it. As I said above, we should add `auto` to fields at some point, though it's a bit less pressing to me than adding implicit arguments to the record type parameters. 

This is because, the `auto` implicit behaviour only affects the generated constructor, and so it is not a lot of hassle to add a 'smart constructor' instead.

src/Idris/Desugar.idr
ohad: They're just a way to help the type-checker see what types are.

I think Idris2 is better about working this out anyway, and would probably have a more direct way to do it.

If `is'`was bound in a `let` instead of `<-`, then I could've moved the `the` statement into the bind. This way of doing it saves me mentioning the monad this is in. (In this case, `Core`, so not a big deal.)

ohad:
PS: thanks for taking a close look at this. I'm still a newb in this area, and the more eyeballs covering my patches, the better.

edwinb:
Thanks for the update and the useful discussion and reviewing! There may be more to tinker with, but this is a useful addition as it is now, I think.

PR #296, opened 2020-04-18
WIP: Bugfix named interfaces

### LINKED ISSUE #291: Named Implementations aren't found by other Implementations with using clause
# Steps to Reproduce

```
module Test

interface MagmaT a where
  op: a -> a -> a

interface MagmaT a => SemigroupT a where
  assoc: (x, y, z: a) -> (x `op` y) `op` z = x `op` (y `op` z)

[NamedMagma] MagmaT Bool where
  False `op` False = False
  _     `op` _     = True

[NamedSemigroup] SemigroupT Bool using NamedMagma where
  assoc = ?hole
```

# Expected Behavior

Typecheck with a hole named `?hole`

# Observed Behavior

```
1/1: Building Test (Test.idr)
Test.idr:7:26--9:1:While processing constructor SemigroupT at Test.idr:6:1--9:1 at Test.idr:6:1--9:1:
Can't find an implementation for MagmaT a
```

fabianhjr:
Closes #291

There seemed to be an issue of what was expected of interface implementations, in particular `name` appears to be a grammar for symbols/operators and not of identifiers.

Additionally something is off with the elaborator but I haven't been able to pin what.

### DIFF
--- a/src/Idris/Desugar.idr
+++ b/src/Idris/Desugar.idr
@@ -690,15 +690,17 @@ mutual
       expandConstraint (Nothing, p)
           = map (\x => (Nothing, x)) (pairToCons p)
 
-  desugarDecl ps (PImplementation fc vis pass is cons tn params impname nusing body)
-      = do is' <- traverse (\ ntm => do tm' <- desugar AnyExpr ps (snd (snd ntm))
+  desugarDecl ps (PImplementation fc vis pass is cons interfaceName params impname nusing body)
+      = do log 10 ("Desugaring implementation of " ++ show interfaceName ++ " named " ++ show impname)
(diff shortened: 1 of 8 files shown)


fabianhjr:
I am having trouble with the interfaceName being lost somewhere.

```
LOG 10: Desugaring implementation of [] named Just MagmaT
LOG 10: Desugaring implementation of [] named Just SemigroupT
```

edwinb:
A `name` could be a symbol or operator, and although it would be a strange thing to do, it is therefore valid for a named implementation to be an operator. So the issue here isn't in the parsing, but rather in when the elaborator adds the named implementations to the searchable set. I have a fix for the issue, which I'll push shortly, so I'm going to close this PR.

PR #299, opened 2020-04-18
Allow spaces in file path for generated Chez programs

christian-public:
If one try to execute an Idris 2 program (`idris2 --exec main Main.idr`) where `Main.idr` is located in a folder with spaces, the generated Chez program will fail to run. Instead it shows an error message like:
```
build/exec/_tmpchez: line 4: /Users/username/Library/Mobile: No such file or directory
```

I added a test where `idris2` is run inside a folder with spaces. (I named the test `chez016` because `chez015` is already taken by #284)

This issue was originally introduced in fd7a2e5.


edwinb:
Thanks for the fix. Something funny has happened with the test numbering though - would you mind updating it, then I'll merge? Ta.

christian-public:
Interestingly, I could resolve the conflict directly from the GitHub page. It is now ready to be merged :)

edwinb:
Splendid, thank you!

PR #305, opened 2020-04-21
Port prelude's inline API documentation from Idris 1

ska80:


edwinb:
Thanks! We'd better make the doc comments meaningful soon too...

PR #309, opened 2020-04-22
Export 'takeBefore' function

ska80:


edwinb:
I'd rather not export this one from the Prelude since it's only there for helping with the range syntax, and is partial, so dosen't really fit with the guidelines for what's in the Prelude. And now I look, I don't think `takeUntil` should be exported either! I think it would be okay for both in Data.Stream though.

edwinb:
In fact I've just made that change (exporting the functions from Data.Stream), so I'll close this PR.

ska80:
> In fact I've just made that change (exporting the functions from Data.Stream), so I'll close this PR.

Thanks!  Makes sense.

PR #310, opened 2020-04-22
Fix builds on *BSD

Kaiepi:
This makes it so gmake gets used instead of make on *BSD and refactors how
platforms are handled in general.


edwinb:
Thanks!

PR #311, opened 2020-04-22
Make a couple improvements for the Racket backend

Kaiepi:
This fixes an error that would get thrown when attempting to use Racket as a backend:
```
/tmp/tmp.0.tSCpq7Hki.rkt:16:42: exact-floor: unbound identifier
  in: exact-floor
  location...:
   /tmp/tmp.0.tSCpq7Hki.rkt:16:42
  context...:
   do-raise-syntax-error
   for-loop
   [repeats 1 more time]
   finish-bodys
   lambda-clause-expander
   loop
   [repeats 4 more times]
   finish-expanding-body
   for-loop
   finish-bodys
   lambda-clause-expander
   for-loop
   loop
   [repeats 7 more times]
   module-begin-k
   expand-module
   ...
```

OpenBSD's Racket package is minimal, so packages that are probably included on other platforms aren't included with it, so this also adds documentation for which ones are needed to run Idris 2.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -77,3 +77,13 @@ After installing chicken scheme you may need to install the 'numbers' package.
 Racket is available from:
 
 + https://download.racket-lang.org/
+
+The following packages are required in order to run Idris:
+
+- base
+- racket-lib
+- compiler-lib
+- r6rs-lib
+- math-lib
+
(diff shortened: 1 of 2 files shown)


edwinb:
Thanks! I haven't paid much attention to the Racket back end, except to test any additions to the libraries...

Issue #307, opened 2020-04-22
Ambigous Implementations when using `using` on a constrainted interface implementation


fabianhjr:
# Steps to Reproduce

```
module Test

interface Simple a where
  op : a -> a -> a

interface Simple a => Constraint a where
  square : a -> a
  square x = x `op` x  

-- And
[NamedSimple1] Simple Bool where
  op True True = True
  op _    _    = False

-- Or
[NamedSimple2] Simple Bool where
  op False False = False
  op _     _     = True

[Constrainted1] Constraint Bool using NamedSimple1 where
[Constrainted2] Constraint Bool using NamedSimple2 where
```

# Expected Behavior

Should typecheck and use their corresponding implementations

# Observed Behavior

```
Test.idr:21:1--23:1:While processing right hand side of Constrainted2 at Test.idr:21:1--23:1:
Multiple solutions found in search. Possible correct results:
	NamedSimple1
	NamedSimple2
```

PR #314, opened 2020-04-23
'using' multiple named instances in scope fix

MarcelineVQ:
Change where the open hints is snapshotted. This is done too late, causing named implementations to remain in scope.
addressing https://github.com/edwinb/Idris2/issues/307
caused by the reordering in https://github.com/edwinb/Idris2/commit/d39c701f280c4ca46f9ca2ba36b59eea3958277d


edwinb:
Ah yes, I missed that, thanks!

PR #332, opened 2020-04-27
Make System.File.readFile keep newlines

ziman:
`System.File.readFile` on a file containing `"foo\nbar"` returns `"foobar"`.

This patch fixes it. I'm not sure about the efficiency but I hope it's not horrible.


edwinb:
Oh, I fixed this without noticing this PR, sorry! I did it the same way.
There's still something not quite right, because if there isn't a newline at the end of the file it shouldn't be in the input that's read. But this at least restores the old behaviour.

PR #333, opened 2020-04-27
[ fix #326 ] Interleaving let binding and declarations

### LINKED ISSUE #326: one type annotation in let means all other definitions must have type annotation
If possible, please include complete self contained source files (for example by uploading to https://gist.github.com) that exhibit the issue in addition to quoting from them here. The smaller the example the better!

# Steps to Reproduce

Check
```
f : a -> b
f a' =
    let x : a
        x = a'
        y = 1
    in ?xyz

g : a -> b
g a' =
    let x : a
        x = a'
        y : Int
        y = 1
    in ?xyz
```

# Expected Behavior

Type checks fine

# Observed Behavior

```
Errors (1)
Test.idr:217:9
While processing right hand side of f at Test.idr:214:1--220:1:
No type declaration for Test.y
```

It can be 'fixed' by having a separate let block for things with type annotations and things without. If this is expected behaviour then:

1. I couldn't find any documentation about it
2. I found it very unintuitive. Indeed it took me a while to work out that I could have an additional block with the unannotated items

gallais:
This is a work in progress. Compilation currently fails with the
following internal error:

idris: Erasure.hs:lamToLet': unexpected input: vs = [{eta_4}], tm = Idris.Parser.case block in case block in mkLets at ./Idris/Parser.idr:602:8-19 at ./Idris/Parser.idr:602:22-26 {e_0} {e_5} {e_6} {e_2} !!V 0!! !!V 0!!
CallStack (from HasCallStack):
  error, called at src/Idris/Erasure.hs:617:27 in idris-1.3.2-0f546d8533795a7031ff67b47abc3f394a048b391a392808377322d50f6ed1f3:Idris.Erasure


ziman:
This seems to happen in overapplied lambdas, which I did not address. The RHS in `Erasure.hs:617` should probably be `mkApp tm vs`.

ziman:
`mkApp tm vs` typechecks but the de Bruijn indices in `vs` will probably be wrong because we're reapplying them under new let bindings. The existing code of `lamToLet` probably suffers from this bug already. :(

This will need a bit more work.

ziman:
https://github.com/idris-lang/Idris-dev/pull/4851 — I'll test it a bit and then remove the draft status.

Issue #297, opened 2020-04-18
Using the arrow keys in the REPL should cycle through recently used commands


malte-v:
Currently using the arrow keys just prints ANSI sequences, which is different from the Idris 1 REPL.

# Steps to Reproduce
Enter the Idris 2 REPL and use the arrow keys.

# Expected Behavior
It cycles through the recently used commands.

# Observed Behavior
Escape sequences get printed.

gallais:
Note that you can use rlwrap to cope with the current limitations

edwinb:
Yes, I run through `rlwrap` (in fact I've aliased `idris2`). This is a fairly big feature request, in practice, requiring readline support or similar, and then leads to wanting tab completion of function names. I actually wonder if a better way is to write a separate fancy REPL via the IDE protocol, and to keep what Idris 2 itself does as simple as possible.

PR #335, opened 2020-04-28
#297: Adding support for readline at the REPL

rgrover:
Switching the REPL to use readline instead of getLine. Added Text/Readline for bindings.

Note: the implementation of readline currently doesn't turn the received input into a buffer managed by the runtime system, and hence is quite likely incomplete.

This is a strawman. I'm happy to take this further and turn the REPL into a separate lightweight application as suggested [here](https://github.com/edwinb/Idris2/issues/297#issuecomment-617094838). I'd need some input for that.


rgrover:
This adds a dependency on readline. The CI system will need to be updated if this is to go ahead.

FranckS:
I would definitely love to see the Repl for idris2 become like the one idris1 had. Especially navigating through recently typed commands with the upper and lower keys, and if doable tab suggestions/auto completion. I don't know what's the best way to do that, so this message is just an encouragement, many thanks for your work on this @rgrover !

edwinb:
We definitely need a nicer REPL, and proper readline support will help there. I just run via `rlwrap` at the moment. In the end, I think a nicer way to do it is to go via the IDE protocol and to keep what Idris itself does as simple as possible - this is at least partly about giving me less to worry about in the core of Idris itself as really I need to understand what's going on in all parts of the system. Also because it already takes ages to build and more features will make it even worse. Getting to a self hosted state will help here, but that'll still take a little while.

So I think my main worry at the moment is that this doesn't do much more than running via `rlwrap` and to take it further will slow things down even more for developers. What input would you need for developing a separate REPL application?

The IDE protocol is described at http://docs.idris-lang.org/en/latest/reference/ide-protocol.html
Idris 2 doesn't do all of it yet, but it does do the basics. Actually you could test it by writing a nice REPL for Idris 1 which Idris 2 would then support later!

PR #337, opened 2020-04-29
Export the implementation of `Traversable (Vect k)`

ziman:
For some reason, it was hidden but it's useful. (I need it.)


edwinb:
Oops! I think when this happens it results from the `%access` directives in Idris 1, but I decided we shouldn't have them in Idris 2 because it's too easy to export things by accident that way.

PR #339, opened 2020-04-29
Make SortedMap foldable and traversable

ziman:
This patch implements `Foldable` and `Traversable` for `SortedMap k`.


edwinb:
Thanks!

Issue #338, opened 2020-04-29
`Nat` too slow in the evaluator


rgrover:
A simple recursion to compute half of a natural number is expensive to execute on Idris2. 

# Steps to Reproduce
Executing `half` (in the following code) scales super-linearly with the argument. 

```idris
import Data.Vect
import Data.Nat -- remove this for Idris1

-- View for dividing a Nat in half; taken literally from libs/base/Data/Nat/Views.idr (Idris1)
data Half : Nat -> Type where
     HalfOdd : {n : Nat} -> Half (S (n + n))
     HalfEven : {n : Nat} -> Half (n + n)

-- Covering function for the `Half` view
half : (n : Nat) -> Half n
half Z = HalfEven {n=Z}
half (S k) with (half k)
  half (S (h + h)) | HalfEven = HalfOdd
  half (S (S (h + h))) | HalfOdd =
    let result = HalfEven {n = S h} in
      rewrite plusSuccRightSucc h h in result
```

# Expected Behavior
The time complexity should be at most linear in `n` (or faster). The same on Idris1 was almost instantaneous.

# Observed Behavior
`half 40` took more than 20 seconds on my laptop.

rgrover:
A straightforward recursive implementation of `half` also takes super-linear time to execute:

```idris
half : (n : Nat) -> Nat
half = go 0
  where
    go : (acc : Nat) -> (x: Nat) -> Nat
    go acc Z = acc
    go acc (S k) =
      case k of
        Z => acc
        S j => go (S acc) j
```

`half 1000` took around 1 second
`half 2000` took around 5 seconds
`half 4000` took around 20 seconds

In this case, the slowdown was also observed with Idris1.

rgrover:
The following code, which transfers one counter into another has the same poor performance as `half` above.

```idris
construct : (n : Nat) -> Nat
construct = go Z
  where
    go : (acc: Nat) -> (x: Nat) -> Nat
    go acc 0 = acc
    go acc (S k) = go (S acc) k
```

This tells me that perhaps there is a quadratic cost for constructing `Nat`s. The `S` operator could be quadratic.

Can someone please point me towards resources I could use to investigate further?

fabianhjr:
[...]

PR #340, opened 2020-04-29
Add noNewtype option to data types

christian-public:
Since the newtype optimisation was added (471a47255fb2757b0efce52560e30a0aa88bdeaa), code generators are no longer able to add special handling of the generated constructor/deconstructor for data types that are affected by the newtype optimisation.

To restore this functionality, I have added a new option (`noNewtype`) to disable the newtype optimisation for individual data types.

### Motivation

In my [Erlang code generator](https://github.com/chrrasmussen/idris2-erlang) I have added [special handling](https://github.com/chrrasmussen/idris2-erlang/blob/66cb4a2d5c00934467484a0694650224a89393f0/idris2/src/Compiler/Erlang/CExp.idr#L125-L133) to some data types, like `ErlAtom`, `ErlBinary` and `ErlCharlist` (from the [`Erlang` module](https://github.com/chrrasmussen/idris2-erlang/blob/66cb4a2d5c00934467484a0694650224a89393f0/idris2/libs/erlang/Erlang.idr#L10-L26)). By doing this I can convert values between an Idris-specific representation to an Erlang-specific representation (and back), like so:

- `MkErlAtom "test"` will generate an Erlang atom: `'test'`
- `MkErlBinary "test"` will generate an Erlang binary: `<<"test"/utf8>>`
- `MkErlCharlist "test"` will generate an Erlang charlist: `"test"`

These data types fulfil the critera for the newtype optimisation, which means that, without the `noNewtype` option, the code generator is unable to add its own special handling.

### Other comments

- The name is supposed to match the existing `noHints` option, but I am happy to change it if anyone has any suggestions.

- The included test is placed in the `chez` section because it rely on printing the internal structure of the data types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Compiler updates:
 * Data types with a single constructor, with a single unerased arguments,
   are translated to just that argument, to save repeated packing and unpacking.
   (c.f. `newtype` in Haskell)
+  + A data type can opt out of this behaviour by specifying `noNewtype` in its
+    options list. `noNewtype` allows code generators to apply special handling
+    to the generated constructor/deconstructor, for a newtype-like data type,
+    that would otherwise be optimised away.
 * 0-multiplicity constructor arguments are now properly erased, not just
   given a placeholder null value.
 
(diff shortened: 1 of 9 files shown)


fabianhjr:
Mhhh, going to try this branch for #338 :o

edwinb:
Good idea, thanks!

PR #343, opened 2020-04-30
Fix Chicken CG after updates in compiler

mokshasoft:
Since Chicken CG is commented out it had gotten out of sync with the rest of the compiler.


mokshasoft:
Removing since Chicken backend is removed.

PR #348, opened 2020-05-03
Implement pragma for declaring nat optimizations

fabianhjr:


fabianhjr:
Helped me improve the performance of an alternative idris2 prelude I am playing with, tested with  mock of ProjectEuler problem #1:

```
module NaiveSolution

import Prelude

import Data.Integral
import Data.Nat.Integral

main : LinkedList Nat
main = filter (`divBy` (the Nat 3)) [(the Nat 1)..100]
```

fabianhjr:
Marceline discovered that this PR causes an explosion of memory usage while compiling Idris2 in Core/Binary (Extra ~ 5GB or so of memory usage)

edwinb:
This does make the nat hack a bit less hacky, thanks! Is the compile time performance still an issue? Sometimes that can be due to ambiguity resolution going a bit out of control.

fabianhjr:
Yes, there is still an issue with Idris1 compilation of Idris2.

Idris2 runs fine though.

src/Idris/Parser.idr
```diff
@@ -1045,6 +1048,66 @@ directive fname indents
          n <- name
          atEnd indents
          pure (PrimChar n)
+
+  -- Prim Nat Optimizations
+
+  <|> do pragma "builtinNatZero"
```
gallais, line 1056: Could we refactor this as something like:

```
do c <- builtinPragma
   n <- name
   atEnd indents
   pure (c n)
```

where `builtinPragma` is a big `choice` of parsers of the form
`PrimNatSucc <$ pragma "builtinNatSucc"`

Issue #349, opened 2020-05-03
find-ipkg doesn't load the ipkg dependencies


berewt:
Dependencies are ignored when you use `--find-ipkg`.

# Steps to Reproduce

`test.ipkg`:

```
package test

depends = contrib
modules = Foo
```

`Foo.idr`:
```
module Foo

import Data.List.TailRec
```

```
$ idris2 --find-ipkg --check Foo.idr
```

# Expected Behavior

It typechecks.

# Observed Behavior

```
   idris2 --find-ipkg --check Foo.idr
(toplevel):1:1--1:1:Data.List.TailRec not found
```

# Comments

it may be enough to load the package in the `findIpkg` method of `Package.idr`. I've tried it locally and it works. I'm going to PR the change.

edwinb:
Closing since @berewt's recent PR fixes this.

PR #350, opened 2020-05-03
--find-pkg load dependencies

berewt:
It addresses issues #349.

It includes all the packages of the `depends` field when it finds the `ipkg` file.


edwinb:
Thanks!

PR #353, opened 2020-05-05
Fix casting Char to String

ska80:


edwinb:
I was about to suggest using `cast` but it seems Idris 1 has the same problem! Thanks for the fix.

PR #354, opened 2020-05-05
Add System.Random

rgrover:
Add system support for generation of random values.


edwinb:
Thanks for this! Just one small thing, though - would you mind putting it in `contrib` instead of `base` for now? I'd like to think a bit more about organisation of effectful things, especially with regard to the interfaces they live in, and if it's in `base` that implies a bit more stability.

(There are some other things in `base` that probably shouldn't be, following this rule. I'll deal with that later...)

rgrover:
I don't intend to add anything further.

edwinb:
Excellent, thanks.

PR #358, opened 2020-05-09
Add System.Clock to fetch current time for various clock types.

rgrover:
Fetches current-time in multiple formats and allows back-ends to offer a subset of the clocks.


edwinb:
Thanks, I've been wanting something like this. At some point we'll need to think about how to convert the foreign calls to C, but probably not for a while.

PR #370, opened 2020-05-12
Refactor makefiles

ska80:


edwinb:
Ah, that does look neater, thanks. It's not immediately obvious what's going wrong with the tests though.

ska80:
> Ah, that does look neater, thanks. It's not immediately obvious what's going wrong with the tests though.

I have noticed the tests started failing after I merged with 9a0608b01f77a3f60dbd04e2ea49d3583bff2316

edwinb:
Looks like I broke something, but I don't know why travis hasn't complained at me... sorry about that! I'll fix it then I'll merge.

ska80:
> Looks like I broke something, but I don't know why travis hasn't complained at me... sorry about that! I'll fix it then I'll merge.

with the latest fixes, now all tests pass

edwinb:
I merged this, but I ran into some problems, and then I wanted to revert for the moment, but that failed: https://github.com/edwinb/Idris2/pull/380

I'll change the `-O2` and the `CC := clang` thing so I can get back to hacking, but I don't know what to to about the `include` thing when I try `make -C dist/rts clean`

edwinb:
I guess the `include` problem is because there's some variables set at the top level. Maybe for what I want adding a top level rule is enough to solve to problem since everything else builds okay.

ska80:
> I merged this, but I ran into some problems, and then I wanted to revert for the moment, but that failed: #380
> 
> I'll change the `-O2` and the `CC := clang` thing so I can get back to hacking, but I don't know what to to about the `include` thing when I try `make -C dist/rts clean`

Fixed in PR #383

Now `make -C dist/rts clean` works as expected.

PR #373, opened 2020-05-13
Idiom brackets syntax was fixed in the docs

buzden:


edwinb:
Oops, thanks for the fix.

PR #374, opened 2020-05-15
Fix tests/chez/chez017 test

ska80:


edwinb:
Oops! I have no idea why I didn't notice I'd done that, or why Travis has stopped emailing me when I do something daft. Thanks for the fix.

PR #380, opened 2020-05-16
Revert "Refactor makefiles"

edwinb:
Reverts edwinb/Idris2#370

This turned out to cause a few problems, and I can't immediately see how to fix them so I'll need to revert for now. They are:

* In the top level `config.mk`, setting `CC ?= clang` doesn't do anything, since `CC` is set by default, so it should probably be `:=`
* Even if `OPT` isn't set, it builds `idris2.c` with `-O2` due to a setting in `dist/Makefile`. Ideally, by default, the code in `dist/rts` should be built with `-O2` since its stable rts code, but during development we don't want `-O2` for the generated `idris2.c`
* Most worryingly, when I do `make -C dist/rts clean` I get an error:

        make: Entering directory '/home/edwin/Research/Yaffle/dist/rts'
        ../config.mk:1: /config.mk: No such file or directory
        make: *** No rule to make target '/config.mk'. Stop.
  
  I'm no `Makefile` expert, as you can probably tell, so I'd rather not get tied up in knots trying to work this one out.


edwinb:
So much for reverting being simple. I guess I'll leave it for now.

ska80:
> * Even if `OPT` isn't set, it builds `idris2.c` with `-O2` due to a setting in `dist/Makefile`. Ideally, by default, the code in `dist/rts` should be built with `-O2` since its stable rts code, but during development we don't want `-O2` for the generated `idris2.c`
> * Most worryingly, when I do `make -C dist/rts clean` I get an error:
>   ```
>     make: Entering directory '/home/edwin/Research/Yaffle/dist/rts'
>     ../config.mk:1: /config.mk: No such file or directory
>     make: *** No rule to make target '/config.mk'. Stop.
>   ```

@edwinb I have fixed those issues in PR #383

Now `make -C dist/rts clean` works as expected.

PR #381, opened 2020-05-16
Add useful functions for List

andylokandy:
Added some trivial functions (basically from Data.Vect and Idris1) to Data.List:

- `last : (l : List a) -> {auto ok : NonEmpty l} -> a`
- `last' : List a -> Maybe a`
- `init : (l : List a) -> {auto ok : NonEmpty l} -> List a`
- `init' : List a -> Maybe (List a)`
- `unfold: (f : a -> Maybe a) -> (x : a) -> List a`
- `foldl1 : (a -> a -> a) -> (l : List a)  -> {auto ok : NonEmpty l} -> a`
- `foldr1 : (a -> a -> a) -> (l : List a)  -> {auto ok : NonEmpty l} -> a`
- `foldl1' : (a -> a -> a) -> List a -> Maybe a`
- `foldr1' : (a -> a -> a) -> List a -> Maybe a`


libs/base/Data/List.idr
```diff
@@ -29,6 +29,15 @@ drop Z     xs      = xs
 drop (S n) []      = []
 drop (S n) (x::xs) = drop n xs
 
+||| Generate a list by repeatedly applying a partial function until exhausted.
+||| @ f the function to iterate
+||| @ x the initial value that will be the head of the list
+public export
+iterate : (f : a -> Maybe a) -> (x : a) -> List a
+iterate f x  = x :: case f x of
```
gallais: This (partial) function is usually called `unfold`.
It morally is a function generating codata (a Colist in this case).

It generalises to other co-datatypes: for a stream you'd have
`(f : a -> a)`; for a binary cotree you'd have `(f : a -> (a, a))`,
etc.

libs/base/Data/List.idr
```diff
@@ -308,6 +333,24 @@ tail' : List a -> Maybe (List a)
...
 tail' (x::xs) = Just xs
 
+||| Attempt to retrieve the last element of a non-empty list.
+|||
+||| If the list is empty, return `Nothing`.
+export
+last' : List a -> Maybe a
+last' [] = Nothing
+last' [x] = Just x
+last' (x::y::ys) = last' (y::ys)
```
gallais: I was expecting the following

```suggestion
last' []          = Nothing
last' xs@(_ :: _) = Just (last xs)
```

Same for `init` and `foldr1`. We already have pure functions doing
the required recursion so why bother repeating ourselves? This also
means that proofs about the primed versions will be able to reuse
results proven for the non-primed ones!

Lastly, it'll be more efficient in the `init` and `foldr1` case as
we won't be bubbling up the `Just` from the base case back to the
outside of the computation.

libs/base/Data/Strings.idr
```diff
@@ -52,7 +43,7 @@ words s = map pack (words' (unpack s))
 ||| ```
 unwords' : List (List Char) -> List Char
 unwords' [] = []
-unwords' ws = assert_total (foldr1 addSpace ws) where
+unwords' (x::xs) = assert_total (foldr1 addSpace (x::xs)) where
```
gallais: I would use an `as` pattern here to make it clear we care about
the *shape* of the data but not about taking it apart and use
its subterms.

So I'd write `unwords' ws@(_ :: _)`.

libs/contrib/Data/String/Extra.idr
```diff
@@ -71,12 +70,12 @@ public export
...
 
-||| Concatenate the strings from a `Foldable` containing strings, separated by
-||| the given string.
+||| Concatenate the strings from a list of strings, separated by
+||| the given seperator.
 public export
-join : (sep : String) -> Foldable t => (xs : t String) -> String
-join sep xs = drop (length sep)
-                   (foldl (\acc, x => acc ++ sep ++ x) "" xs)
+join : (sep : String) -> List String -> String
```
gallais, line 76: This is a breaking change. Do we really want the less general
type signature? What does it bring us?

libs/base/Data/List.idr
andylokandy: Good point. Now that I merely reused the name in Data.Steam, shall we aslo change it?

libs/base/Data/List.idr
andylokandy: Same to me. I chose the repeated version just because the other primed functions in Data.List do so. I'm not familiar with the convention in idris, so which style is preferred?

Edit: What I mentioned is Data.List of Idris1.

libs/contrib/Data/String/Extra.idr
andylokandy: This change is to remove the complexity of dropping the heading separator after foldl.

Further, this new interface sticks tighter to other functions in the same module, since most of the String functions work with List, and join was the only one that uses Foldable.

libs/base/Data/List.idr
andylokandy: Renamed: 
- `Data.List.iterate` => `Data.List.unfold`
- `Data.Stream.iterate` => `Data.Stream.unfold`

PR #385, opened 2020-05-17
Add System.Path

andylokandy:
Add a cross-platform path helper, which handles different formats depending on the platform. It can also be used to append paths, to update the extension and to extract the parent directory.

Hopefully, it can be used to self-host idris2 and, as a result, improves robustness on different platforms.

Based on #381 #377


libs/contrib/System/Path.idr
```diff
@@ -161,7 +159,7 @@ startWith b p = b `elem` (unfold parent p)
 ||| If the path is a normal file, this is the file name. If it's the
 ||| path of a directory, this is the directory name.
 |||
-||| Reutrns Nothing if the final body is ".." or ".".
+||| Reutrns Nothing if the final body is ".." or "."
```
keram:
```suggestion
||| Returns Nothing if the final body is ".." or "."
```

edwinb:
This looks useful to have, thanks. Rather than merging now, though, I want to move over to working with Idris2-SH, so please could you send the PR there instead? I don't plan to keep the repositories in sync very long.

Also - it would be better if this patch was just about System.Path, and if there's refactorings of the list libraries they can be done separately. Mostly the additions there do seem good to have. I'd rather not change names of functions from Idris 1 to Idris 2 (even if they arguably do have better names like `unfold`) especially if they mean even more changes to code in the book.

andylokandy:
Sounds great, I'll migrate this PR to idris2sh soon.

PR #387, opened 2020-05-19
Create new mapError utility function and refactor some parser code

fabianhjr:
This corresponds with edwinb/Idris2-SH#6

### DIFF
--- a/src/Parser/Support.idr
+++ b/src/Parser/Support.idr
@@ -7,6 +7,7 @@ import public Text.Parser
 
 import Core.TT
 import Data.List.Views
+import Utils.Either
 
 %default total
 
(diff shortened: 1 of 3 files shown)


src/Parser/Support.idr
```diff
@@ -46,23 +47,19 @@ eoi
     isEOI EndInput = True
     isEOI _ = False
 
+export
+mapParseError : ParseError (TokenData Token) -> ParseError
```
gallais, line 51: This type annotation is pretty puzzling. May be worth adding a comment.
It's also a bit perverse to call this `map` when the input & output
structures are not the same. Maybe `errorAtNextToken` or something of
the sort?

src/Utils/Either.idr
```diff
@@ -0,0 +1,5 @@
+module Utils.Either
+
```
gallais, line 2: Argh, it seems we have two concurrent hierarchies for these type of useful additional
little functions: `Data.X.Extra` and `Utils.X`. :/

fabianhjr: I didn't notice such duplication of structure. I am inclined to keep it in `Utils.Either` since it is more discoverable than `Data.Either.Extra` but will follow whatever convention @edwinb prefers.

src/Parser/Support.idr
fabianhjr: `toGenericParseError`?

`ParseError (TokenData Token)` is a specialized error type and `ParseError` is a more generic parsing error. I was also a bit conflicted about this naming.

ziman: I think this is a good example of a case where explicit namespace qualification would be useful.

PR #389, opened 2020-05-19
Build standalone executable via gambit backend

andylokandy:
In this PR, the gambit backend statically links foreign libraries (e.g. `libidris2_support.a`) and builds a standalone executable.

Note that if you've installed gambit-c via apt, it'll dynamically link the gambit runtime, so you have to distribute `libgambit.so.4.8.8` (or with another version) as well. Only if you install gambit via source code, you can build the truly standalone executable.

The binary size of a simple program that prints a line to console is about 8MB. In addition, the execution time (unoptimized) is about 0.2s, while chez is about 0.1s.

### DIFF
--- a/src/Compiler/Scheme/Gambit.idr
+++ b/src/Compiler/Scheme/Gambit.idr
@@ -24,13 +24,13 @@ import System.Info
 findGSI : IO String
 findGSI =
   do env <- getEnv "GAMBIT_GSI"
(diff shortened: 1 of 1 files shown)


edwinb:
Ah, thanks, I have been wanting to try more with the gambit CG but after I switched to using C for support code I didn't investigate how to fix this...

I'll merge the other PRs in a bit - this one is nice and easy though :).

PR #390, opened 2020-05-21
Fix building the latest self-hosted Idris 2

ska80:
* Add System.Path to the 'contrib' package
* Update Data.List in the 'base' package

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 
 *.dSYM
 
-/idris2
+/idris2boot
 /runtests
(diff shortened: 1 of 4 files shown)


edwinb:
Oh, I missed this problem, cheers. I don't know if there's an easy way to automatically check that this still works for bootstrapping. I guess we'll just have to check manually every so often unless someone has some ideas.

melted:
It should be pretty easy to make a travis job for it.

